PR #5, opened 2024-08-23
Parsing

dhirenmathur:


app/alembic/env.py
```diff
@@ -21,6 +21,7 @@
 
 # Construct the database URL from environment variables
 POSTGRES_SERVER = os.getenv("POSTGRES_SERVER", "localhost")
+print(POSTGRES_SERVER)
```
vineetshar: kindly remove this

app/core/config.py
```diff
@@ -0,0 +1,11 @@
+from dotenv import load_dotenv
+import os
+
+load_dotenv()
+
+neo4j_config = {
```
vineetshar: can we make this a config provider class please

app/core/crud_utils.py
```diff
@@ -0,0 +1,81 @@
+from sqlalchemy.orm import Session
+from sqlalchemy import and_
+from app.modules.users.user_model import User
+from app.modules.projects.projects_model import Project
+from datetime import datetime
+
+# User CRUD operations
```
vineetshar: I am not aligned with the structure of this file, has scope of improvements 

1. user , projects related stuff sits here , instead of their respective modules 
2. this is not a class. 

The objective of this code feels similar to what can sit in a services file.

app/core/database.py
```diff
@@ -34,3 +34,15 @@ def get_db():
         yield db
     finally:
         db.close()
+
+
+
+class SessionManager:
```
vineetshar: why is this needed? we already use get_db() to provide sessions and let them flow via router via dependency injection, example 

 async def get_conversation_info(
        conversation_id: str,
        db: Session = Depends(get_db)
    ):
        controller = ConversationController(db)
        return await controller.get_conversation_info(conversation_id)

app/firebase_setup.py
```diff
@@ -0,0 +1,25 @@
+import os
+import firebase_admin
+from firebase_admin import auth, credentials
+import base64
+import json
+import logging
+
+def firebase_init():
```
vineetshar: class please

app/main.py
```diff
@@ -67,7 +74,14 @@ def run(self):
         self.add_health_check()
         return self.app
 
+if os.getenv("isDevelopmentMode") == "enabled":
+
```
vineetshar: we already have this code in main 

    def setup_data(self):

        # Setup dummy user and project during application startup
        dummy_data_setup = DummyDataSetup()
        dummy_data_setup.setup_dummy_user()
        dummy_data_setup.setup_dummy_project()
        
        can you fix the conditions for firebase setup please

app/models.py
```diff
@@ -0,0 +1,19 @@
+from sqlalchemy.orm import relationship
```
vineetshar: this file is not needed, we removed this in my pr before merging, please remove and test.

app/modules/intelligence/memory/postgres_history_manager.py
```diff
@@ -0,0 +1,42 @@
+from sqlalchemy.orm import Session
+from typing import List, Optional
```
vineetshar: redundant, was renamed to chat_history_service, please compare to main and check

app/modules/key_management/secret_manager.py
```diff
@@ -0,0 +1,140 @@
...
+from fastapi import Depends, HTTPException
+from google.cloud import secretmanager
+from pydantic import BaseModel, validator
+from app.modules.utils.APIRouter import APIRouter
+
+from app.modules.auth.auth_service import AuthService
+from app.core.mongo_manager import MongoManager
+
+router = APIRouter()
+
```
vineetshar, line 13: it should be structured in a class

app/modules/utils/ai_helper.py
```diff
@@ -0,0 +1,39 @@
+from portkey_ai import createHeaders, PORTKEY_GATEWAY_URL
+from langchain_openai.chat_models import ChatOpenAI
+import os
+from app.modules.key_management.secret_manager import  get_secret
+from app.core.mongo_manager import MongoDBHelper
+
+def get_llm_client(user_id, model_name):
```
vineetshar: can we have this part of a class as well

app/modules/utils/model_helper.py
```diff
@@ -0,0 +1,33 @@
+from sqlalchemy.orm import class_mapper
+from sqlalchemy.orm.exc import DetachedInstanceError
+
+def model_to_dict(model, max_depth=1, current_depth=0):
```
vineetshar: class

app/firebase_setup.py
```diff
@@ -0,0 +1,25 @@
+import os
+import firebase_admin
```
vineetshar: also this file should sit inside utils module

PR #10, opened 2024-08-26
Store explanations from files 

dhirenmathur:


app/modules/parsing/graph_construction/parsing_router.py
```diff
@@ -88,19 +56,39 @@ async def parse_directory(
...
             )
 
             extracted_dir, project_id = await parse_helper.setup_project_directory(
                 repo, repo_details.branch_name, auth, repo, user_id, project_id
             )
 
-            await ParsingService.analyze_directory(
-                extracted_dir, project_id, user_id, db
-            )
+            # Update Celery task invocations
```
vineetshar: not a fan of dead code going into the project

app/modules/parsing/graph_construction/parsing_helper.py
```diff
@@ -4,9 +4,11 @@
 import os
 import tarfile
```
vineetshar, line 4: this file has become too big. Huge cognitive load - please split into multiple files

PR #13, opened 2024-08-27
Project routes and Github Route

GodReaper:
Finished adding project routes 
Finished adding github routes
Added mongodb image to docker compose
update env.template
Fixed mongo ssl error when running in development mode

### DIFF
--- a/app/core/mongo_manager.py
+++ b/app/core/mongo_manager.py
@@ -37,14 +37,25 @@ class MongoManager:
 
             # Establish the connection based on the environment
             if env in ["production", "staging"]:
-                connect(db_name, host=mongo_uri, tlsCAFile=certifi.where())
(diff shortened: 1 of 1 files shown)


app/core/mongo_manager.py
```diff
@@ -23,19 +23,38 @@ def __init__(self):
...
+            raise
+
     def _connect(self):
         if self._client is None:
             try:
-                mongodb_uri = os.environ.get("MONGO_URI")
-                if not mongodb_uri:
-                    raise ValueError("MONGO_URI environment variable is not set")
-
-                self._client = MongoClient(
```
vineetshar, line 33: why did we remove connection pooling?

GodReaper: Ahh mistake on my end I will put it back

PR #15, opened 2024-08-30
[DEV-222, DEV-258] Support public repo and trigger email on completion 

GodReaper:
Added Email sending Support and Github Repo fetching from public url

### DIFF
--- a/app/modules/parsing/graph_construction/parsing_helper.py
+++ b/app/modules/parsing/graph_construction/parsing_helper.py
@@ -29,6 +29,7 @@ class ParseHelper:
     def __init__(self, db_session: Session):
         self.project_manager = ProjectService(db_session)
         self.db = db_session
+
     @staticmethod
     async def clone_or_copy_repository(
         repo_details: RepoDetails, db: Session, user_id: str
@@ -55,19 +56,21 @@ class ParseHelper:
                 github = Github()
                 repo = github.get_repo(repo_details.repo_name)
             except Exception as public_repo_error:
-                logging.error(f"Failed to fetch public repository: {str(public_repo_error)}")
-                
+                logging.error(
+                    f"Failed to fetch public repository: {str(public_repo_error)}"
+                )
+
                 # If public repo fetch fails, try private repo
                 try:
                     response, auth, owner = github_service.get_github_repo_details(
                         repo_details.repo_name
                     )
-                    
+
                     if response.status_code != 200:
                         raise HTTPException(
                             status_code=400, detail="Failed to get installation ID"
                         )
-                    
+
                     app_auth = auth.get_installation_auth(response.json()["id"])
                     github = Github(auth=app_auth)
                     repo = github.get_repo(repo_details.repo_name)
@@ -75,7 +78,9 @@ class ParseHelper:
                     if isinstance(private_repo_error, HTTPException):
                         raise private_repo_error
                     else:
-                        logging.error(f"Failed to fetch private repository: {str(private_repo_error)}")
+                        logging.error(
+                            f"Failed to fetch private repository: {str(private_repo_error)}"
+                        )
                         raise HTTPException(
                             status_code=404, detail="Repository not found on GitHub"
                         )


app/modules/parsing/graph_construction/parsing_service.py
```diff
@@ -70,6 +71,8 @@ async def parse_directory(
             await project_manager.update_project_status(
                 project_id, ProjectStatusEnum.READY
             )
+            if(ProjectStatusEnum.READY):
+                await EmailHelper().send_email(user_email)
```
dhirenmathur: use asyncio.create task so that any error in email sending does not make the API error out. 

I had asked you to back this by an env property as well transactionn_emails_enabled=True

app/modules/parsing/graph_construction/parsing_helper.py
```diff
@@ -66,10 +77,13 @@ async def download_and_extract_tarball(
...
         try:
             tarball_url = repo_details.get_archive_link("tarball", branch)
+            headers = {}
+            if auth is not None:
+                headers["Authorization"] = f"Bearer {auth.token}"
             response = requests.get(
                 tarball_url,
                 stream=True,
-                headers={"Authorization": f"{auth.token}"},
+                headers=headers,
```
dhirenmathur: this is very different to the previous version: headers={"Authorization": f"{auth.token}"}, 

why do we need to add Bearer here, is this backward compatible?

app/modules/parsing/graph_construction/parsing_helper.py
```diff
@@ -45,19 +45,30 @@ async def clone_or_copy_repository(
...
+                        status_code=400, detail="Failed to get installation ID"
+                    )
+                app_auth = auth.get_installation_auth(response.json()["id"])
+                github = Github(auth=app_auth)
+            except HTTPException:
+                # Handling public repository
+                response, owner = github_service.get_public_github_repo(
+                    repo_details.repo_name
+                )
+                github = Github()  # No authentication for public repos
```
dhirenmathur: handle this only for 404 error not all exceptions

GodReaper: Ok

app/modules/github/github_service.py
```diff
@@ -125,6 +125,7 @@ def get_repos_for_user(user):
 
             for installation in installations:
                 app_auth = auth.get_installation_auth(installation["id"])
+                github = Github(auth=app_auth)
```
vineetshar, line 128: why is this variable needed?

GodReaper: This fetches the auth from github removing this causes a exception to trigger

app/modules/utils/email_helper.py
```diff
@@ -1,18 +1,26 @@
...
+            os.environ.get("TRANSACTION_EMAIL_ENABLED", "False").lower() == "true"
+        )
+        resend.api_key = self.api_key
 
     async def send_email(self, to_address):
+        if not self.transaction_emails_enabled:
+            return
+
         params = {
             "from": "raj@momentum.sh",
```
dhirenmathur: 1) Keep from address as a env param too
2) Can we avoid  HTML and instead use simple Text emails.

app/modules/parsing/graph_construction/parsing_helper.py
```diff
@@ -79,7 +79,7 @@ async def download_and_extract_tarball(
             tarball_url = repo_details.get_archive_link("tarball", branch)
             headers = {}
             if auth is not None:
-                headers["Authorization"] = f"Bearer {auth.token}"
+                headers = {"Authorization": f"{auth.token}"}
```
dhirenmathur, line 102: without bearer added it works for both public and private?

app/modules/parsing/graph_construction/parsing_helper.py
```diff
@@ -56,19 +57,18 @@ async def clone_or_copy_repository(
...
+            except HTTPException:
+                try:
+                    response, owner = github_service.get_public_github_repo(
+                        repo_details.repo_name
+                    )
+                    github = Github()
+                    repo = github.get_repo(repo_details.repo_name)
+                except Exception:
+                    raise HTTPException(
+                        status_code=404, detail="Repository not found on GitHub"
```
dhirenmathur: I meant handle the 404 from the request not send a 404 , altough that is a good catch which i missed, so keep it. 

if request = 404: 
  #handle public repo

app/modules/parsing/graph_construction/parsing_helper.py
GodReaper: Yes

app/modules/utils/email_helper.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class EmailHelper:
+    def __init__(self):
+        self.api_key = os.environ.get("RESEND_API_KEY")
+        self.transaction_emails_enabled = (
+            os.environ.get("TRANSACTION_EMAILS_ENABLED", "false").lower() == "true"
+        )
+        self.from_address = os.environ.get("EMAIL_FROM_ADDRESS", "example@momentum.sh")
+        resend.api_key = self.api_key
```
vineetshar, line 13: example@momentum.sh in case of bugs will look extremely bad, make it support@momentum.sh

GodReaper: Ok

PR #24, opened 2024-09-05
(DEV-262) Secret management support for openai and anthropic with migration to sql

vineetshar:
Screenshot 2024-09-05 at 3 56 24 PM
Screenshot 2024-09-05 at 3 56 29 PM

Didnt upload rest of the screenshots as they had our openai key haha

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,6 +1,16 @@
 #!/bin/bash
 source .env
 
+# Set up Service Account Credentials
+export GOOGLE_APPLICATION_CREDENTIALS="./service-account.json"
+
+# Check if the credentials file exists
+if [ ! -f "$GOOGLE_APPLICATION_CREDENTIALS" ]; then
+    echo "Service Account Credentials file not found at $GOOGLE_APPLICATION_CREDENTIALS"
+    echo "Please ensure the service-account.json file is in the current directory."
+    exit 1
+fi
+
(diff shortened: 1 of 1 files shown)


app/alembic/versions/20240905144257_342902c88262_add_user_preferences_table.py
```diff
@@ -0,0 +1,43 @@
...
+        sa.Column("user_id", sa.String(), nullable=False),
+        sa.Column("preferences", sa.JSON(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["user_id"],
+            ["users.uid"],
+        ),
+        sa.PrimaryKeyConstraint("user_id"),
+    )
+    op.create_index(
+        "idx_user_preferences_user_id", "user_preferences", ["user_id"], unique=False
```
dhirenmathur: how do you index a json and what are the benefits?

vineetshar: we are not indexing the json, it's the user id being indexed which will speed up lookups, joins, and foreign key operations involving the user_preferences table.

dhirenmathur: my bad

PR #27, opened 2024-09-09
Tools for code

vineetshar:
Screenshot 2024-09-09 at 8 59 32 AM
Screenshot 2024-09-09 at 9 03 38 AM
Screenshot 2024-09-09 at 9 43 42 AM

### DIFF
--- a/app/celery/tasks/parsing_tasks.py
+++ b/app/celery/tasks/parsing_tasks.py
@@ -35,21 +35,6 @@ def process_parsing(
 ) -> None:
     logger.info(f"Task received: Starting parsing process for project {project_id}")
     try:
-        # Check if the project is already being processed
-        active_tasks = celery_app.control.inspect().active()
-        if active_tasks:
-            for worker, tasks in active_tasks.items():
-                for task in tasks:
-                    task_args = eval(task["argsrepr"])
-                    if (
-                        task["name"] == "app.celery.tasks.parsing_tasks.process_parsing"
-                        and task_args[3] == project_id
-                    ):
-                        logger.info(
-                            f"Project {project_id} is already being processed by task {task['id']}. Skipping task."
-                        )
-                        return
-
         parsing_service = ParsingService(self.db)
         asyncio.run(
             parsing_service.parse_directory(


app/celery/celery_app.py
```diff
@@ -0,0 +1,55 @@
...
+from celery import Celery
+
+from app.core.models import *  # This will import and initialize all models
+
+# Redis configuration
+redishost = os.getenv("REDISHOST", "localhost")
+redisport = int(os.getenv("REDISPORT", 6379))
+redisuser = os.getenv("REDISUSER", "")
+redispassword = os.getenv("REDISPASSWORD", "")
+queue_name = os.getenv("CELERY_QUEUE_NAME", "staging")
```
dhirenmathur, line 13: can we do this in the config.py file that has neo4j creds

app/modules/conversations/conversation/conversation_service.py
```diff
@@ -299,6 +293,7 @@ async def _generate_and_stream_ai_response(
             )
 
         try:
+            print(f"Running agent {conversation.project_ids[0]} with query: {query}")
```
dhirenmathur: logger

app/modules/github/github_service.py
```diff
@@ -220,3 +220,55 @@ def get_public_github_repo(repo_name: str):
...
+
+            file_contents = repo.get_contents(file_path)
+
+            if isinstance(file_contents, list):
+                raise HTTPException(
+                    status_code=400, detail="Provided path is a directory, not a file"
+                )
+
+            decoded_content = file_contents.decoded_content.decode("utf-8")
+            lines = decoded_content.split("\n")
```
dhirenmathur: handle non-text files

app/celery/celery_app.py
vineetshar: Ideally no: 
1. They currently get picked from .env which makes sense to me. Adding another layer doesnt.
2. Bigger problem : Right now celery module can be copied as a microservice and doesnt have a lot of dependency. Adding a config layer in between will tightly bind it.

app/modules/github/github_service.py
```diff
@@ -33,8 +31,8 @@ def get_github_repo_details(repo_name: str):
         auth = AppAuth(app_id=app_id, private_key=private_key)
         jwt = auth.create_jwt()
         owner = repo_name.split("/")[0]
-        repo = repo_name.split("/")[1]
-        url = f"https://api.github.com/repos/{owner}/{repo}/installation"
+
+        url = f"https://api.github.com/repos/{owner}/{repo_name.split('/')[1]}/installation"
```
dhirenmathur, line 35: can we handle local repos in a follwo up PR

app/modules/github/github_service.py
```diff
@@ -204,71 +207,25 @@ def get_branch_list(repo_name: str):
...
-                app_auth = auth.get_installation_auth(response.json()["id"])
-                github = Github(auth=app_auth)
-                repo = github.get_repo(repo_name)
-
-            file_contents = repo.get_contents(file_path)
-
-            if isinstance(file_contents, list):
+                return github, repo
+            except Exception as private_error:
+                logger.error(f"Failed to access private repo: {str(private_error)}")
```
dhirenmathur, line 227: Follow up PR : can we flip this, do private first and then public

vineetshar: sure

app/modules/github/github_service.py
vineetshar: sure

PR #31, opened 2024-09-13
Agent Integrations 

dhirenmathur:
Contains changes for: 
1) Citations in separate key in response 
2) Node ids 
3) New knowledge graph integration 
4) Unit Test agent 
5) Agentic RAG over KG

### DIFF
--- a/app/modules/intelligence/tools/kg_based_tools/get_code_from_node_name_tool.py
+++ b/app/modules/intelligence/tools/kg_based_tools/get_code_from_node_name_tool.py
@@ -1,13 +1,18 @@
-from typing import Dict, Any
-from sqlalchemy.orm import Session
+from typing import Any, Dict
+
 from neo4j import GraphDatabase
-from app.modules.github.github_service import GithubService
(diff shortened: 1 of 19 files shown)


app/modules/intelligence/tools/kg_based_tools/get_code_from_node_name_tool.py
```diff
@@ -0,0 +1,93 @@
+from typing import Dict, Any
+from sqlalchemy.orm import Session
+from neo4j import GraphDatabase
+from app.modules.github.github_service import GithubService
+from app.core.config_provider import config_provider
+from app.modules.projects.projects_model import Project
+
+class GetCodeFromNodeNameTool:
```
vineetshar, line 11: are these duplicate tools? changes to existing tool?

dhirenmathur: no does not look like duplicate or changes, not sure why they show up as new files in this PR
image

PR #32, opened 2024-09-13
Knowledge graph for flow infrmation

dhirenmathur:
Contains changes to add flow level inferences to every entry node in the graph


app/modules/parsing/graph_construction/parsing_helper.py
```diff
@@ -365,7 +363,7 @@ async def check_commit_status(self, project_id: str) -> bool:
             )
             return False
 
-        if len(repo_name.split("/")) >= 2:
+        if len(repo_name.split("/")) < 2:
```
vineetshar, line 366: why this change and what's the impact?

PR #49, opened 2024-09-17
Posthog integration

GodReaper:
Screenshot 2024-09-15 202426

### DIFF
--- a/app/modules/parsing/graph_construction/parsing_controller.py
+++ b/app/modules/parsing/graph_construction/parsing_controller.py
@@ -44,7 +44,7 @@ class ParsingController:
 
                 logger.info(f"Submitting parsing task for new project {new_project_id}")
                 await project_manager.update_project_status(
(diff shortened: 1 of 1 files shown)


app/celery/tasks/parsing_tasks.py
```diff
@@ -41,15 +38,25 @@ def process_parsing(
     logger.info(f"Task received: Starting parsing process for project {project_id}")
     try:
         parsing_service = ParsingService(self.db, user_id)
-        asyncio.run(
-            parsing_service.parse_directory(
+
+        async def run_parsing():
```
vineetshar, line 42: @dhirenmathur do we want to log parsing completion time in server logs?

GodReaper: I think i left because i was testing the repos yesterday i forgot to revert this file

dhirenmathur: Let it be in server logs

PR #52, opened 2024-09-18
Conversation listing improvements

vineetshar:
Send project details when fetching all conversations of a user
Screenshot 2024-09-18 at 5 28 14 PM

### DIFF
--- a/app/modules/users/user_service.py
+++ b/app/modules/users/user_service.py
@@ -1,15 +1,14 @@
 import logging
 from datetime import datetime
-from typing import List, Set
+from typing import List
 
 from sqlalchemy import desc
 from sqlalchemy.exc import SQLAlchemyError
-from sqlalchemy.orm import Session, joinedload
+from sqlalchemy.orm import Session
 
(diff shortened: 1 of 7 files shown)


app/modules/users/user_service.py
```diff
@@ -115,17 +84,46 @@ def get_user_by_uid(self, uid: str):
...
+                .limit(limit)
+                .all()
+            )
+            
+            # Fetch projects for all conversations in a single query
+            project_ids = set()
+            for conversation in conversations:
+                project_ids.update(conversation.project_ids)
+            
+            projects = self.db.query(Project).filter(Project.id.in_(project_ids)).all()
```
dhirenmathur: we can leverage sqlalchemy relationship here right?

dhirenmathur (changes requested):

app/modules/users/user_service.py
vineetshar: done

dhirenmathur (approved):

PR #53, opened 2024-09-18
Agent updates 

dhirenmathur:


app/celery/tasks/parsing_tasks.py
```diff
@@ -59,18 +59,17 @@ async def run_parsing():
...
                 f"Parsing process took {elapsed_time:.2f} seconds for project {project_id}"
             )
 
         asyncio.run(run_parsing())
-
         logger.info(f"Parsing process completed for project {project_id}")
     except Exception as e:
         logger.error(f"Error during parsing for project {project_id}: {str(e)}")
-        raise  # Let the automatic retry handle it
+        raise self.retry(exc=e, countdown=60, max_retries=3)
```
vineetshar: this is redundant, just a normal raise would retry.

app/modules/intelligence/tools/change_detection/change_detection.py
```diff
@@ -247,7 +247,7 @@ async def get_code_changes(self, project_id):
                         }
 
                     entry_points = InferenceService(
-                        self.sql_db
+                        self.sql_db, "dummy"
```
vineetshar, line 250: why dummy Is hardcoded here?

app/modules/parsing/knowledge_graph/query_router.py
```diff
@@ -15,7 +15,7 @@
 
 @router.post("/query", response_model=List[QueryResponse])
 async def query_vector_index(request: QueryRequest, db: Session = Depends(get_db)):
-    inference_service = InferenceService(db)
+    inference_service = InferenceService(db, "dummy")
```
vineetshar, line 18: same, why is dummy hardcoded here

PR #55, opened 2024-09-19
posthog and sentry disabled in dev env

GodReaper:


app/modules/utils/posthog_helper.py
```diff
@@ -5,18 +5,24 @@
 
 class PostHogClient:
     def __init__(self):
+        self.is_development = os.getenv("ENV") == "development"
```
vineetshar: the problem with this approach is that it will still require potshot env in local system setup and initialise it, it will just not send the event

app/main.py
```diff
@@ -48,11 +48,12 @@ def __init__(self):
         self.include_routers()
 
     def setup_sentry(self):
-        sentry_sdk.init(
-            dsn=os.getenv("SENTRY_DSN"),
-            traces_sample_rate=1.0,
-            profiles_sample_rate=1.0,
-        )
+        if os.getenv("ENV") != "development":
```
vineetshar: in both the places, lets change the conditions to enable sentry and posthog in staging and production envs. It's because naming those two are in our control, and the local one people can name whatever they want, dev, development, local etc.

app/modules/utils/posthog_helper.py
```diff
@@ -1,28 +1,36 @@
...
+            self.api_key = os.getenv("POSTHOG_API_KEY")
+            self.posthog_host = os.getenv("POSTHOG_HOST")
+            self.posthog = Posthog(self.api_key, host=self.posthog_host)
+        else:
+            self.posthog = None
 
     def send_event(self, user_id: str, event_name: str, properties: dict):
         """
-        Sends a custom event to PostHog.
+        Sends a custom event to PostHog unless the environment is set to development.
```
vineetshar: in the stage and production env

PR #81, opened 2024-09-26
Updates to knowledge graph creation (node cleanup and docstring generation)

GodReaper:
Screenshot 2024-09-26 at 9 31 55 PM

### DIFF
--- a/app/modules/intelligence/agents/agentic_tools/rag_agent.py
+++ b/app/modules/intelligence/agents/agentic_tools/rag_agent.py
@@ -82,9 +82,9 @@ class RAGAgent:
 
             ## Step 1: Initial Context Retrieval
             - If user provided node IDs are present:
-            1. Use the "Get Code and docstring From Node ID" tool for each provided node ID.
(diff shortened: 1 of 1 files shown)


app/modules/parsing/knowledge_graph/inference_service.py
```diff
@@ -500,6 +502,18 @@ async def update_neo4j_with_docstrings(
...
+
+            if len(repo_name.split("/")) < 2:
+                return  # Do not cleanup if it's a local repo
+
+            session.run(
+                """
+                MATCH (n:NODE {repoId: $repo_id})
+                REMOVE n.text, n.signature
+                """, repo_id=repo_id,
+            )
```
dhirenmathur: can we do this in the query on line 495 instead of doing a new write query again?

dhirenmathur: use separate query based on local repo or not. 

if local : 
existing query 
else: 
old query + remove in same query

PR #87, opened 2024-09-28
Regenerate

dhirenmathur:
Without node ids (still needs to contain empty body)
image
with node ids: 
image
empty node ids : 
image


app/celery/celery_app.py
```diff
@@ -55,14 +55,13 @@ def configure_celery(queue_prefix: str):
...
 
 configure_celery(queue_name)
 
+# Import the lock decorator
+
 # Import tasks to ensure they are registered
 import app.celery.tasks.parsing_tasks  # noqa # Ensure the task module is imported
-
-# Import the lock decorator
-from celery.contrib.abortable import AbortableTask
```
vineetshar, line 68: add this back with noqa please

PR #104, opened 2024-10-02
Auth for tools

dhirenmathur:
image
image
image
image
image
image


app/modules/github/github_service.py
```diff
@@ -113,10 +118,10 @@ def get_file_content(
             decoded_content = content_bytes.decode(encoding)
             lines = decoded_content.splitlines()
 
-            if start_line == end_line == 0:
+            if (start_line == end_line == 0) or (start_line == end_line == None):
                 return decoded_content
 
-            selected_lines = lines[start_line:end_line]
+            selected_lines = lines[start_line - 2 : end_line]
```
vineetshar, line 124: why -2?

app/modules/intelligence/agents/agents_service.py
```diff
@@ -39,3 +39,13 @@ async def list_available_agents(self) -> List[AgentInfo]:
                 description="An agent specialized in generating detailed analysis of code changes in your current branch compared to default branch. Works best with Py, JS, TS",
             ),
         ]
+
+    def cleanup_citations(self, citations: List[str]) -> List[str]:
```
vineetshar: want to understand the intent of this function

dhirenmathur: just to format the citations, without this the citations are of the format: 

repo-branch-uid/projects/root_dir/path/to file 
after this it is 
root_dir/path/to file

app/modules/github/github_service.py
dhirenmathur: This is super hacky: 
There's an issue with parsing where the start line is 1 later than it should be. 
I made it -1 to match function definition
and made it -2 to include decorators etc, space in worst case if nothing else is there

vineetshar: can we add this as a comment , so that it anyone who reads the code knows why this is happening.

app/modules/intelligence/agents/agents_service.py
vineetshar: can we rename to format_citations then, cleanup can be understood wrongly like it means maybe a database cleanup of citations.

vineetshar (approved):
LGTM

PR #137, opened 2024-10-28
Share chat public

GodReaper:


GodReaper:
Screenshot 2024-10-29 at 8 58 59 PM
Screenshot 2024-10-29 at 8 59 03 PM

app/alembic/versions/20241026151704_e824fba2b86a_new_migration.py
```diff
@@ -0,0 +1,30 @@
+"""New migration
+
+Revision ID: 20241026151704_e824fba2b86a
```
dhirenmathur: is this file still needed? just undo this migration and generate a new one instead of fixing this in the next one

dhirenmathur (changes requested):

app/modules/conversations/conversations_router.py
```diff
@@ -160,11 +160,21 @@ async def share_chat(
...
             message="Chat shared successfully!", sharedID=shared_conversation
         )
     except ShareChatServiceError as e:
         raise HTTPException(status_code=400, detail=str(e))
+
+@router.get("/conversations/{conversation_id}/shared-emails", response_model=List[str])
+async def get_shared_emails(
+    conversation_id: str,
+    db: Session = Depends(get_db),
+):
```
dhirenmathur, line 178: Where is the authorisation? 
Add auth and make sure that the get emails endpoint returns 401 when a user who is not the conversation owner tries to request them

dhirenmathur (changes requested):

PR #177, opened 2024-11-14
feat: public user profile pic

Yash-pede:


start.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 
-source .env
+# source .env
```
dhirenmathur:
```suggestion
source .env
```

app/modules/users/user_service.py
```diff
@@ -143,3 +146,14 @@ def get_user_ids_by_emails(self, emails: List[str]) -> List[str]:
         except Exception as e:
             logger.error(f"Error fetching user ID by emails {emails}: {e}")
             return None
+    def get_user_profile_pic(self, uid: str) -> UserProfileResponse:
```
dhirenmathur:
```suggestion


    def get_user_profile_pic(self, uid: str) -> UserProfileResponse:

Comment (dhirenmathur) [reply app/modules/users/user_service.py]: make this function and the internal api call async @Yash-pede

**Review by dhirenmathur — DISMISSED** [review]

File: app/modules/users/user_service.py [hunk 3]
```diff
@@ -143,3 +147,16 @@ def get_user_ids_by_emails(self, emails: List[str]) -> List[str]:
         except Exception as e:
             logger.error(f"Error fetching user ID by emails {emails}: {e}")
             return None
+
+
+    async def get_user_profile_pic(self, uid: str) -> UserProfileResponse:
+        try:
+            user_record = await asyncio.to_thread(auth.get_user, uid)
```
vineetshar, line 154: seems very untraditional way of doing it, why are we doing it this way @Yash-pede ?

vineetshar:
Generally looks, good, left one comment needing clarification @Yash-pede 

I was also wondering if firebase will always return a url even if the user has not uploaded their picture for example? Have we tested this scenario?

app/modules/users/user_service.py
Yash-pede: auth.getuser is not a async function and also to unblock the main thread until it is fetching the user details 
I have showed it @dhirenmathur

Yash-pede: Will discuss with him Should i remove it ??

vineetshar (approved):

PR #192, opened 2024-11-25
feat: initial changes for support for isDevelopmentMode=enabled

waveywaves:
- Skip firebase init in dev mode.
- Add logging for AuthService in dev mode
- Disable PortKey initialization in dev mode

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -33,7 +33,6 @@ logging.basicConfig(
     level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"
 )
 
-
 class MainApp:
     def __init__(self):
         load_dotenv(override=True)
@@ -64,8 +63,11 @@ class MainApp:
         )
 
     def setup_data(self):
-        if os.getenv("ENV") != "development":
+        if os.getenv("ENV") == "development":
+            logging.info("Development mode enabled. Skipping Firebase setup.")
+        else:
             FirebaseSetup.firebase_init()
+
 
     def initialize_database(self):
         # Initialize database tables
--- a/app/main.py
+++ b/app/main.py
@@ -36,6 +36,9 @@ logging.basicConfig(
 class MainApp:
     def __init__(self):
         load_dotenv(override=True)
+        if os.getenv("isDevelopmentMode") == "enabled":
+            if os.getenv("ENV") != "development":
+                logging.error("Development mode enabled but ENV is not set to development. Exiting.")
         self.setup_sentry()
         self.app = FastAPI()
         self.setup_cors()
--- a/app/main.py
+++ b/app/main.py
@@ -49,7 +49,7 @@ class MainApp:
         self.include_routers()
 
     def setup_sentry(self):
-        if os.getenv("ENV") != "development":
+        if os.getenv("ENV") == "production":
             sentry_sdk.init(
                 dsn=os.getenv("SENTRY_DSN"),
                 traces_sample_rate=0.25,
--- a/app/main.py
+++ b/app/main.py
@@ -39,6 +39,7 @@ class MainApp:
         if os.getenv("isDevelopmentMode") == "enabled":
             if os.getenv("ENV") != "development":
                 logging.error("Development mode enabled but ENV is not set to development. Exiting.")
+                exit(1)
         self.setup_sentry()
         self.app = FastAPI()
         self.setup_cors()
--- a/app/main.py
+++ b/app/main.py
@@ -36,10 +36,9 @@ logging.basicConfig(
 class MainApp:
     def __init__(self):
         load_dotenv(override=True)
-        if os.getenv("isDevelopmentMode") == "enabled":
-            if os.getenv("ENV") != "development":
-                logging.error("Development mode enabled but ENV is not set to development. Exiting.")
-                exit(1)
+        if os.getenv("isDevelopmentMode") == "enabled" and os.getenv("ENV") != "development":
+            logging.error("Development mode enabled but ENV is not set to development. Exiting.")
+            exit(1)
         self.setup_sentry()
         self.app = FastAPI()
         self.setup_cors()
--- a/app/main.py
+++ b/app/main.py
@@ -46,7 +46,7 @@ class MainApp:
         self.include_routers()
 
     def setup_sentry(self):
-        if os.getenv("ENV") == "production":
+        if os.getenv("ENV") != "development":
             sentry_sdk.init(
                 dsn=os.getenv("SENTRY_DSN"),
                 traces_sample_rate=0.25,
--- a/app/main.py
+++ b/app/main.py
@@ -42,10 +42,7 @@ class MainApp:
         self.setup_cors()
         self.initialize_database()
         self.check_and_set_env_vars()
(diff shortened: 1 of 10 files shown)


vineetshar:
Hi @waveywaves some thoughts towards this pull request. 

The separation between ENV and isDevelopmentMode is intentional in its naming and use case. 

The ENV signifies the environment the app is running in, signfying configuration to load in dev / stage or prod env. 

The isDevelopmentMode is a bit different in required use case, it tells the app that it can run without dependencies such as firebase, GitHub configuration etc. 

The ENV development still needs firebase / gcp / GitHub etc because it simply is about running potpie's backend locally. But isDevelopmentMode disables all these to bring in support for lrunning without most dependencies and support local parsing for example, and other things in future. 

You might need to redo some stuff in the PR for this.

waveywaves:
Hey @vineetshar, ah okay. Thanks for helping me out with that. I'll update this info in the contributing doc in that case if it isn't there.

waveywaves:
@vineetshar I have updated the PR, lmkwyt

app/modules/utils/posthog_helper.py
```diff
@@ -11,12 +11,12 @@ def __init__(self):
         self.environment = os.getenv("ENV")
 
         # Only initialize PostHog if not in development mode
-        if self.environment == "production":
+        if self.environment == "development":
```
vineetshar: no this will enable posthog outside prod in envs like stage. We want posthog only working prod.

app/main.py
```diff
@@ -33,23 +33,23 @@
...
         self.check_and_set_env_vars()
-        if os.getenv("isDevelopmentMode") == "enabled":
-            self.setup_data()
-        else:
-            FirebaseSetup.firebase_init()
+        self.setup_data()
         self.include_routers()
 
     def setup_sentry(self):
-        if os.getenv("ENV") == "production":
```
vineetshar: same problem here, removing this enables sentry in envs like stage as well. Should be prod only

app/main.py
```diff
@@ -33,23 +33,23 @@
     level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s"
 )
 
-
 class MainApp:
     def __init__(self):
         load_dotenv(override=True)
+        if os.getenv("isDevelopmentMode") == "enabled":
```
vineetshar: correct condition but readability can be improved by using and instead of inner if

app/modules/intelligence/provider/provider_service.py
```diff
@@ -83,7 +86,12 @@ def get_large_llm(self):
             else "openai"
         )
 
-        if preferred_provider == "openai":
+        if self.environment != "development":
```
vineetshar: I do not understand the motive and replacement of this action. what happens if we disable this llm?

app/modules/utils/posthog_helper.py
waveywaves: I will revert this change then

app/modules/intelligence/provider/provider_service.py
waveywaves: I have added this check as the first check during setting the development mode. Here we are skipping LLM initialization if Development mode is not enabled. The statement I have used here is wrong. I should be checking `if os.getenv("isDevelopmentMode") == enabled`.

app/main.py
waveywaves: this was an embarrassing mistake, my bad

waveywaves:
@vineetshar the PR is ready for review. Thank you in advance.

app/modules/intelligence/provider/provider_service.py
```diff
@@ -83,7 +85,12 @@ def get_large_llm(self):
             else "openai"
         )
 
-        if preferred_provider == "openai":
+        if os.getenv("isDevelopmentMode") != "enabled":
```
dhirenmathur: Only portkey shouldn't be initialised, but LLM client should be otherwise ai features won't work

app/main.py
```diff
@@ -66,6 +65,12 @@ def setup_cors(self):
             allow_headers=["*"],
         )
 
+    def setup_data(self):
+        if os.getenv("isDevelopmentMode") == "enabled":
+            logging.info("Development mode enabled. Skipping Firebase setup.")
```
dhirenmathur, line 70: @vineetshar was this supposed to include sunny user creation?

app/modules/utils/posthog_helper.py
```diff
@@ -26,7 +26,9 @@ def send_event(self, user_id: str, event_name: str, properties: dict):
             event_name (str): The name of the event to track.
             properties (dict): Additional properties related to the event.
         """
-        if self.environment != "production":
+        if self.environment == "development":
```
dhirenmathur: Let this stay production as we don't want staging to emit posthog events

dhirenmathur (dismissed):

app/modules/intelligence/provider/provider_service.py
waveywaves: Noted. Updating the code to not initialize portkey during LLM client init.

app/modules/utils/posthog_helper.py
waveywaves: reverting this change

waveywaves:
@dhirenmathur thank you for looking into this, updated based on your review

app/main.py
vineetshar: ideally yes, since we disable GitHub auth, and parsing and further interactions with the project does require a user , so we should have a default user creation to actually let things happen further.

So @waveywaves maybe what should happen is that all the major apis auth should become optional, and if development mode is enabled, we should skip auth and use the default user. This default user should ideally be inserted in database (if not present) when app starts with development mode.

We do not need to go to each router and make auth optional but do it at a common place. I think this is critical to go forward.

@waveywaves let me know if you have questions

vineetshar:
> @dhirenmathur thank you for looking into this, updated based on your review

@waveywaves one feedback left about user creation, can we do that as well please.

waveywaves:
@vineetshar yes, shall revert to the comment asap

app/main.py
waveywaves: > ideally yes, since we disable GitHub auth, and parsing and further interactions with the project does require a user , so we should have a default user creation to actually let things happen further.

yep, I see the default user being returned by the AuthService

> We do not need to go to each router and make auth optional but do it at a common place. I think this is critical to go forward.

check_auth which is used by all the routers is already returning a default user.

waveywaves: I haven't really made any code changes to setup_data, I have just put the firebase init validation for isDevelopmentMode in the setup_data function. The AuthService already returns the default_user with check_auth for `ifDevelopmentMode=enabled`. What more changes do we need to make exactly ?

vineetshar: yes I just saw that, that part has not been tested though. I think I will resolve this conversation and test the rest of pr out for approval.

vineetshar:
Approving this since disabling the mentioned things work as part of this PR. But development mode will be fully functional once we support parsing (which is to be done today 28th Nov), before that it is not really usable but is outside the scope of this specific pull request.

vineetshar (approved):
LGTM

Issue #279, opened 2025-02-24
Changes Required while using Azure open AI api key.


Dimpal-Kalita:
Hi, I am using the AZURE open ai API key for gpt-4. normal applications, it require the following arguments
```py
from langchain.chat_models import AzureChatOpenAI 

llm = AzureChatOpenAI(
    openai_api_base=AZURE_ENDPOINT,
    openai_api_version=AZURE_API_VERSION,
    openai_api_key=AZURE_API_KEY,
    deployment_name= AZURE_DEPLOYMENT_NAME,
    openai_api_type="azure"
)
```

In local development, I have changed the ```MODEL_CONFIGS``` in ```provider_service.py```
``` py
  MODEL_CONFIGS = {
        "openai": {
            "small": {
                "crewai": {"model": "openai/gpt-4o-mini"},
                "langchain": {
                    # "model": "gpt-4o-mini",
                    "class": AzureChatOpenAI,
                    "openai_api_base": os.getenv("AZURE_ENDPOINT"),
                    "openai_api_version": os.getenv("AZURE_API_VERSION"),
                    "openai_api_key": os.getenv("AZURE_API_KEY"),
                    "deployment_name": os.getenv("AZURE_DEPLOYMENT_NAME"),
                    "openai_api_type": "azure",
                },
            },
            "large": {
                "crewai": {"model": "openai/gpt-4o"},
                "langchain": {
                    # "model": "gpt-4o",
                    "class": AzureChatOpenAI,
                    "openai_api_base": os.getenv("AZURE_ENDPOINT"),
                    "openai_api_version": os.getenv("AZURE_API_VERSION"),
                    "openai_api_key": os.getenv("AZURE_API_KEY"),
                    "deployment_name": os.getenv("AZURE_DEPLOYMENT_NAME"),
                    "openai_api_type": "azure",
                },
            },
        },
```

But, I am getting the following error, while trying to make conversations using the following curl,
```bash
curl --location 'http://localhost:8001/api/v1/conversations/<my id>/message/' \
--header 'Content-Type: application/json' \
--data '{
    "content": "Your question or request here"
  }'
```

```
RROR:app.modules.conversations.conversation.conversation_service:Failed to generate and stream AI response for conversation 019536f6-df6d-7698-ae5f-7e6bc22adb2b: Connection error.
Traceback (most recent call last):
[...]

PR #288, opened 2025-02-26
Adding Support for Azure openai API keys

Dimpal-Kalita:
Added  ```AZURE_API_BASE=```  and  ```AZURE_API_VERSION=``` in environment variables as they are necessary for interacting with the model.
Issue #279 

## Summary by CodeRabbit

- **New Features**
  - Introduced new configuration options to enhance Azure API integration.
  - Enabled dynamic assignment of API endpoint and version settings based on environment configurations.
  - Added new environment variables: `LLM_API_BASE` and `LLM_API_VERSION` for improved customization.

### DIFF
--- a/app/modules/intelligence/provider/provider_service.py
+++ b/app/modules/intelligence/provider/provider_service.py
@@ -315,8 +315,14 @@ class ProviderService:
             # ollama + portkey is not supported currently
             extra_params["base_url"] = PORTKEY_GATEWAY_URL
             extra_params["extra_headers"] = createHeaders(
-                api_key=self.portkey_api_key, provider=routing_provider
+                api_key=self.portkey_api_key,
+                provider=routing_provider,
+                custom_host=os.environ.get("LLM_API_BASE"),
+                api_version=os.environ.get("LLM_API_VERSION"),
             )
+        elif provider == "azure":
+            extra_params["api_base"] = os.environ.get("LLM_API_BASE")
+            extra_params["api_version"] = os.environ.get("LLM_API_VERSION")
 
         try:
             if stream:
@@ -369,8 +375,14 @@ class ProviderService:
             # ollama + portkey is not supported currently
             extra_params["base_url"] = PORTKEY_GATEWAY_URL
             extra_params["extra_headers"] = createHeaders(
-                api_key=self.portkey_api_key, provider=routing_provider
+                api_key=self.portkey_api_key,
+                provider=routing_provider,
+                custom_host=os.environ.get("LLM_API_BASE"),
+                api_version=os.environ.get("LLM_API_VERSION"),
             )
+        elif provider == "azure":
+            extra_params["api_base"] = os.environ.get("LLM_API_BASE")
+            extra_params["api_version"] = os.environ.get("LLM_API_VERSION")
 
         try:
             if provider == "ollama":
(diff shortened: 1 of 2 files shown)


dhirenmathur (changes requested):
Thank you so much, @Dimpal-Kalita, for opening a PR!

A few observations:
	1.	Environment Keys
Can we keep it generic with environment keys like LLM_API_BASE and LLM_API_VERSION? If these keys are present, they should be added to the request.
	2.	LLM Client Initialization
There are three places where an LLM client is being prepared. You have handled two of them—in call_llm and call_llm_with_structured_output—but the LLM client is also initialized inside the _initialize_llm function (the CrewAI LLM client). This should be updated to include the new parameters as well.
	3.	Portkey Integration
Portkey integration also needs to be updated. In the case of Portkey, the API_BASE is the Portkey URL, and the Azure URL needs to be passed through Portkey headers as custom_host (I might be mistaken on this, so please double-check).

Dimpal-Kalita:
I faced an issue because of mostprobably LiteLLM. I couldn't figure out the reason for the error, but I was able to bypass it by naming the environment variable ``AZURE_API_BASE ``.
The issue is similar to this [Link](https://github.com/langchain-ai/langchain/discussions/27010)
It is being bypassed by renaming the environment variable because of this code in ```litellm/main.py```
``` py
            api_base = api_base or litellm.api_base or get_secret("AZURE_API_BASE")
```

There are multiple iterations of querying the LLM is being done right. The ```api_base``` is getting passed for the first two iterations, then it disappears. I couldn't figure out why it is behaving like that. The above fix was able to bypass that.

Dimpal-Kalita:
@dhirenmathur Suggestion 1 and suggestion 2 are handled. For the portkey ```api_base``` and azure ```api_base``` conflict, I see no option except passing it as ```custom_host```. 

I have tested the other changes it is working as expected.

dhirenmathur:
Awesome! I will review and test it out today!

dhirenmathur:
@nndn can you review/test this please

nndn:
@Dimpal-Kalita can you share an example of what all env variables we need to set to work with azure openai?

Dimpal-Kalita:
```
LLM_PROVIDER=azure
LLM_API_KEY= xxxxxxxx
LOW_REASONING_MODEL=azure/model
HIGH_REASONING_MODEL=azure/model
LLM_API_BASE=xxxxx
LLM_API_VERSION= xxxxx
```

app/modules/intelligence/provider/provider_service.py
```diff
@@ -309,6 +309,9 @@ async def call_llm(
             extra_params["extra_headers"] = createHeaders(
                 api_key=self.portkey_api_key, provider=routing_provider
             )
+        if provider == "azure":
+            extra_params["api_base"] = os.environ.get("LLM_API_BASE")
+            extra_params["api_version"] = os.environ.get("LLM_API_VERSION")
```
nndn:
```suggestion
            extra_params["extra_headers"] = createHeaders(
                api_key=self.portkey_api_key,
                provider=routing_provider,
                custom_host=os.environ.get("LLM_API_BASE"),
            )
        elif provider == "azure":
            extra_params["api_base"] = os.environ.get("LLM_API_BASE")
            extra_params["api_version"] = os.environ.get("LLM_API_VERSION")
```

nndn: I think it should be something like this, this worked for me. Can you confirm the same? @Dimpal-Kalita

Dimpal-Kalita: API version willn't be passed in this case right to the portkey?
I guess we should add that to the extra params as well.

Something like this,
```py
            extra_params["extra_headers"] = createHeaders(
                api_key=self.portkey_api_key,
                provider=routing_provider,
                custom_host=os.environ.get("LLM_API_BASE"),
            )
            extra_params["api_version"] = os.environ.get("LLM_API_VERSION")
        elif provider == "azure":
            extra_params["api_base"] = os.environ.get("LLM_API_BASE")
            extra_params["api_version"] = os.environ.get("LLM_API_VERSION")
```

nndn: correct, but those headers which you need to send to the llm we need to pass through extra_params field when using portkey. I've also made it into elif so that we aren't passing it twice. Please update the pr and test it out. I'll test it out from my end too once you push changes

nndn: even api_version you need to pass in extra_headers. Also this same thing will be repeated in 2 more places

nndn:
I've left one comment, other than that it looks good. Please merge the latest commits from main, I'll approve it once that's done

Dimpal-Kalita:
I have merged the latest commits from the main but I don't see any comments

Dimpal-Kalita:
@nndn

nndn:
> @nndn 

Can you fix the return type for the new method. It returns a tuple now right?

nndn (approved):

Dimpal-Kalita:
@nndn  Is resolving the comments from CodeRabbit needed?

nndn:
Yup there's only one comment right, mark it as resolved

Dimpal-Kalita:
@nndn is there anything left in this PR?

nndn (approved):

nndn:
No @Dimpal-Kalita . I'll give one last round of testing and merge it off

PR #317, opened 2025-03-20
custom-agent-id-fix

kinshuksinghbist:
## Summary by CodeRabbit

- **New Features**
  - Enhanced conversation details now display more comprehensive agent information for improved context.
  - Updated user conversation views deliver clearer agent role displays based on refined data handling.

- **Chores**
  - Refreshed the UI component dependency to its latest reference for improved stability and compatibility.


app/modules/users/user_controller.py
```diff
@@ -28,6 +30,12 @@ async def get_conversations_for_user(
             repo_name = projects[0].repo_name
             branch_name = projects[0].branch_name
 
+            agent_id = conversation.agent_ids[0] if conversation.agent_ids else None
+
+            custom_agent = self.sql_db.query(CustomAgent).filter_by(id=agent_id).first()
+            display_agent_id = (
+                custom_agent.role if custom_agent else conversation.agent_ids[0]
+            )
             response.append(
```
dhirenmathur: this wil do 100 db calls for 100 conversations, update this logic to do a single db call outside the loop

app/modules/conversations/conversation/conversation_service.py
```diff
@@ -664,6 +665,17 @@ async def get_conversation_info(
...
+            agent_id = conversation.agent_ids[0] if conversation.agent_ids else None
+
+            agent_ids = conversation.agent_ids
+            if agent_id:
+                custom_agent = (
+                    self.sql_db.query(CustomAgent).filter_by(id=agent_id).first()
+                )
+                if custom_agent:
+                    agent_ids = [custom_agent.role]
+
```
dhirenmathur: is it possible to do a join here and get this done in a single query ?

app/modules/users/user_controller.py
dhirenmathur: is it possible to do a join between conversation and agent here and get this done in a single query ?

dhirenmathur (changes requested):

app/modules/conversations/conversation/conversation_service.py
```diff
@@ -664,6 +665,17 @@ async def get_conversation_info(
                 .filter_by(conversation_id=conversation_id, status=MessageStatus.ACTIVE)
                 .count()
             )
+
+            agent_id = conversation.agent_ids[0] if conversation.agent_ids else None
+
+            agent_ids = conversation.agent_ids
+            if agent_id:
+                custom_agent = (
+                    self.sql_db.query(CustomAgent).filter_by(id=agent_id).first()
```
dhirenmathur: can we not check if the agent id is a system id before querying? That would avoid so many db calls

dhirenmathur (dismissed):

PR #322, opened 2025-03-25
fix: structure output

nndn:
## Summary by CodeRabbit

- **Refactor**
	- Optimized internal prompt processing logic for improved backend integration and type handling while maintaining consistent overall functionality. 
	- Removed the `size` parameter and introduced a new `config_type` parameter in the prompt enhancement process.

### DIFF
--- a/app/modules/intelligence/prompts/prompt_service.py
+++ b/app/modules/intelligence/prompts/prompt_service.py
@@ -388,7 +388,8 @@ class PromptService:
         try:
             provider_service = ProviderService(self.db, user["user_id"])
             result = await provider_service.call_llm_with_structured_output(
-                messages=messages, output_schema=EnhancedPromptResponse, size="large"
+                messages=messages,
+                output_schema=EnhancedPromptResponse,  # type: ignore
             )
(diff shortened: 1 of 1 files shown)


app/modules/intelligence/prompts/prompt_service.py
```diff
@@ -388,7 +388,8 @@ async def enhance_prompt(
         try:
             provider_service = ProviderService(self.db, user["user_id"])
             result = await provider_service.call_llm_with_structured_output(
-                messages=messages, output_schema=EnhancedPromptResponse, size="large"
+                messages=messages,
+                output_schema=EnhancedPromptResponse,  # type: ignore
```
dhirenmathur, line 392: can we add config_type = chat or is that default

dhirenmathur (changes requested):

app/modules/intelligence/prompts/prompt_service.py
nndn: I think that's the default

dhirenmathur (approved):

PR #392, opened 2025-04-26
Fix deduplication logic in get_user_repos to preserve repo order

### LINKED ISSUE #355: Fix Duplicate Repository Deduplication Logic in get_user_repos()
**Description:**

The current deduplication logic in the `get_user_repos()` function from `github_router.py` iterates through the list of repositories in reverse order. This approach may not maintain the original order of the repositories, potentially leading to confusion when the data is presented to users. We need to update the logic to preserve the original order while removing duplicates.

**Proposed Changes:**

1. **Modify Deduplication Logic**:
   - Refactor the loop that removes duplicates to iterate in a way that maintains the original order of repositories.
   - Instead of iterating in reverse, iterate normally and use a set to track seen repositories to ensure duplicates are effectively removed while preserving their sequence.

2. **Code Sample Before and After**:
   - Provide a code snippet showing the current implementation and proposed changes for clarity.

**Impacted Files**:
- `app/modules/code_provider/github/github_router.py`: This file contains the `get_user_repos()` function which requires modification.

**Next Steps**: If approved, I will implement the necessary changes to the deduplication logic and submit a pull request for review.

andoriyaprashant:
## Description:
This PR updates the deduplication logic in the get_user_repos() function to preserve the original order of repositories. The previous implementation iterated in reverse, causing the order to be flipped and potentially confusing users.

Changes Made:

- Replaced reverse iteration with a forward loop

- Used repo["full_name"] as the unique identifier

- Ensured the first occurrence of each repo is kept

### DIFF
--- a/app/modules/code_provider/github/github_router.py
+++ b/app/modules/code_provider/github/github_router.py
@@ -21,10 +21,9 @@ async def get_user_repos(
     # Remove duplicates while preserving order
     seen = set()
     deduped_repos = []
-    for repo in reversed(user_repo_list["repositories"]):
-        # Create tuple of values to use as hash key
+    for repo in user_repo_list["repositories"]:
         repo_key = repo["full_name"]
-
+        
         if repo_key not in seen:
             seen.add(repo_key)
             deduped_repos.append(repo)


dhirenmathur:
@nndn issue #355 might be inavalid?

PR #407, opened 2025-05-22
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

Note that your Jenkins might start "more correctly" after this fix ? ;)

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,.codespellrc
+check-hidden = true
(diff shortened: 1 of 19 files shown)


kshitij2040:
Thanks for the contribution here. We didn’t end up picking this up at the time due to product priorities, and it’s now gone stale. I’m going to close this for now. Please feel free to open a new PR if you'd like to take another shot, we’d love to see you contribute to Potpie again.

PR #530, opened 2025-12-12
fix(auth): enforce strict dev mode check to prevent auth bypass in production

shanthanu47:
### Security Fix: Enforce Strict Auth in Dev Mode

**The Vulnerability**: 
Previously, setting `isDevelopmentMode=enabled` would bypass authentication entirely. If this flag were accidentally left on in production, it would be a major security risk.

**The Fix**: 
I added a second required environment variable: `ALLOW_MOCK_AUTH=true`. 
Mock authentication (bypass) is now ONLY allowed if **both** `isDevelopmentMode=enabled` AND `ALLOW_MOCK_AUTH=true` are present.

**Verification**: 
Verified with a script that sets `isDevelopmentMode=enabled` (without the second flag) and confirmed that access is correctly DENIED.

### DIFF
--- a/app/modules/auth/auth_service.py
+++ b/app/modules/auth/auth_service.py
@@ -57,13 +57,23 @@ class AuthService:
         logging.info(f"DEBUG: Credential provided: {credential is not None}")
 
         # Check if the application is in debug mode
-        if os.getenv("isDevelopmentMode") == "enabled" and credential is None:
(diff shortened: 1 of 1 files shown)


kshitij2040:
Thanks for the contribution here. We didn’t end up picking this up at the time due to product priorities, and it’s now gone stale. I’m going to close this for now. Please feel free to open a new PR if you'd like to take another shot, we’d love to see you contribute to Potpie again

PR #551, opened 2025-12-29
feat: Add work email validation for SSO authentication

yashkrishan:
- Block new users with generic email providers (Gmail, Yahoo, Outlook, etc.)
- Block all new GitHub signups (existing users can still sign in)
- Allow legacy users with generic emails to continue using the service
- Add server-side validation in /sso/login and /signup endpoints
- Add email domain validation utilities (extract_domain, is_generic_email)
- Return 403 Forbidden with clear error messages for blocked attempts

## Summary by CodeRabbit

* **New Features**
  * Signups via Google SSO using common personal/generic email domains are now blocked for new accounts (existing legacy accounts with such emails remain allowed).
  * New GitHub-based signups are blocked unless the GitHub provider is already linked to an existing account.
  * Blocked registration attempts return clear 403 error responses with explanatory payloads.

* **Chores**
  * Improved registrable-domain parsing for more reliable generic-email detection.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/app/modules/auth/auth_router.py
+++ b/app/modules/auth/auth_router.py
@@ -35,15 +35,7 @@ from app.modules.auth.unified_auth_service import (
 from app.modules.users.user_service import UserService
 from app.modules.utils.APIRouter import APIRouter
 from app.modules.utils.posthog_helper import PostHogClient
-
-# Import publicsuffix2 for proper domain extraction (handles multi-part TLDs)
-try:
-    from publicsuffix2 import get_sld
-
-    HAS_PUBLICSUFFIX2 = True
-except ImportError:
-    HAS_PUBLICSUFFIX2 = False
-    get_sld = None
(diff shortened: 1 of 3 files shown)


app/modules/auth/auth_router.py
```diff
@@ -36,11 +36,160 @@
...
+except ImportError:
+    HAS_PUBLICSUFFIX2 = False
+    get_sld = None
+
 SLACK_WEBHOOK_URL = os.getenv("SLACK_WEBHOOK_URL", None)
 
 auth_router = APIRouter()
 load_dotenv(override=True)
 
+# Blocked email domains (generic/personal email providers)
```
dhirenmathur: no library we can use for this instead? if not please move this to the config file

dhirenmathur (changes requested):

dhirenmathur (approved):

Issue #348, opened 2025-04-14
Unclear error when accessing private repository branches without permissions


dhirenmathur:
### Description
The `get_branch_list` function in `github_service.py` doesn't properly handle the case where a user tries to access a private repository without the necessary permissions. 

Currently, when a user attempts to fetch branches from a private repository they don't have access to, the function returns a generic 404 error with "Repository not found or error fetching branches" message. This doesn't distinguish between repositories that don't exist and those that exist but are private and inaccessible.

### Expected Behavior
The function should detect private repositories that the user doesn't have access to and return a more specific error message, such as a 403 Forbidden with a clear message that the repository exists but the user doesn't have sufficient permissions.

### Current Behavior
The function returns a generic 404 error for both non-existent repositories and private repositories without access.

### Code Location
`app/modules/code_provider/github/github_service.py` - `get_branch_list` method

### Suggested Fix
Modify the `get_branch_list` method to catch specific GitHub API exceptions that indicate permission issues and return appropriate 403 responses with clear error messages.

Dsantra92:
QOL improvement and not part of #794 migration plan. Closing this for now.

PR #617, opened 2026-02-22
fix: return HTTP 403 with clear message when accessing private repo branches without permissions

Aiudadadadf:
## Summary

get_branch_list in github_service.py was catching GithubException (from the PyGithub library) inside the generic except Exception handler, causing any GitHub API 401/403 response to be re-raised as an HTTP 404 with a misleading "Repository not found or error fetching branches" message.

This made it impossible to tell from the response whether a repository doesn't exist or whether the user simply lacks access.

## Changes

- Added an explicit except GithubException as ge: block **before** the generic except Exception handler
- When ge.status is 401 or 403, raise HTTP 403 with a clear "Access denied" message explaining the user lacks permissions to list branches
- Other GithubException status codes still fall through to a 404, now using ge.data.get('message', ...) for a more descriptive detail string
- Added logger.warning for permission-denied cases (instead of logger.error) since this is expected user error, not a server fault

## Testing

- Manually traced the exception path through get_branch_list for the GitHub repository branch (the else block at line 1022)
- GithubException is already imported at the top of the file (rom github.GithubException import GithubException)
- No new dependencies introduced

Fixes potpie-ai/potpie#348

## Release Notes

* **Bug Fixes**
  * Improved error messages for GitHub repository access issues
  * More accurate HTTP status codes when encountering permission or repository errors during branch retrieval

### DIFF
--- a/app/modules/code_provider/github/github_service.py
+++ b/app/modules/code_provider/github/github_service.py
@@ -1030,6 +1030,26 @@ class GithubService:
                 return {"branches": [default_branch] + branch_list}
         except HTTPException as he:
             raise he
+        except GithubException as ge:
+            if ge.status in (401, 403):
+                logger.warning(
+                    f"Permission denied accessing branches for repo {repo_name}: {ge.data.get('message', str(ge))}"
+                )
+                raise HTTPException(
+                    status_code=403,
+                    detail=(
+                        f"Access denied: repository '{repo_name}' exists but you do not have "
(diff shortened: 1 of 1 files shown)


kshitij2040:
Thanks for the contribution here. We didn’t end up picking this up at the time due to product priorities, and it’s now gone stale. I’m going to close this for now. Please feel free to open a new PR if you'd like to take another shot, we’d love to see you contribute to Potpie again.

PR #635, opened 2026-02-24
fix: Replace Response() with JSONResponse() in signup and custom-token endpoints

### LINKED ISSUE #634: [Bug] Signup endpoint returns JSON body with text/plain content-type header
### Bug

In `app/modules/auth/auth_router.py`, the `POST /signup` and `POST /auth/custom-token` endpoints use `Response(content=json.dumps(...))` to return JSON data. The `Response` class defaults to `Content-Type: text/plain`, so clients that check the content-type header will not parse the body as JSON.

```python
# Current (broken — Content-Type: text/plain):
return Response(content=json.dumps({"error": "Missing uid"}), status_code=400)

# Should be (Content-Type: application/json):
return JSONResponse(content={"error": "Missing uid"}, status_code=400)
```

18 instances across the signup handler and custom-token handler are affected. Meanwhile, other routes in the **same file** (login, SSO login, provider endpoints) correctly use `JSONResponse`.

### Impact

- API clients and frontend code that checks the `Content-Type` header will not parse signup/custom-token responses as JSON
- Inconsistent response format within the same router — login returns `application/json` but signup returns `text/plain` for equivalent JSON bodies

### Expected behavior

All endpoints returning JSON data should use `JSONResponse` (or set `media_type="application/json"`) so the `Content-Type` header is correct.

rohansx:
## Summary

  - Replace 18 instances of `Response(content=json.dumps(...))` with `JSONResponse(content=...)`
    in the signup and custom-token endpoints
  - Remove unused `Response` import

  Fixes #634

  ## Problem

  The `POST /signup` and `POST /auth/custom-token` endpoints used `Response(content=json.dumps(...))`
  which defaults to `Content-Type: text/plain`. Clients checking the content-type header would not
  parse these responses as JSON, despite the body being valid JSON.

  Other endpoints in the same file (login, SSO, provider management) already used `JSONResponse` correctly.

  ## Changes

  - `app/modules/auth/auth_router.py` — replaced all `Response(content=json.dumps(X), status_code=Y)`
    with `JSONResponse(content=X, status_code=Y)`, removed unused `Response` import

## Summary by CodeRabbit

* **Refactor**
  * Improved authentication response handling to use cleaner and more consistent JSON formatting across signup and custom-token flows.

### DIFF
--- a/app/modules/auth/auth_router.py
+++ b/app/modules/auth/auth_router.py
@@ -5,7 +5,7 @@ import os
 import requests
 from dotenv import load_dotenv
 from fastapi import Depends, Request
-from fastapi.responses import JSONResponse, Response
+from fastapi.responses import JSONResponse
 from fastapi.exceptions import HTTPException
 
 from sqlalchemy.orm import Session
@@ -129,12 +129,12 @@ class AuthAPI:
 
         # Validate
         if not uid:
-            return Response(
-                content=json.dumps({"error": "Missing uid"}), status_code=400
+            return JSONResponse(
+                content={"error": "Missing uid"}, status_code=400
             )
         if not email:
-            return Response(
-                content=json.dumps({"error": "Missing email"}), status_code=400
+            return JSONResponse(
+                content={"error": "Missing email"}, status_code=400
             )
 
         user_service = UserService(db)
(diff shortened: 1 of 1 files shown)


rohansx:
@dhirenmathur please take a look

PR #653, opened 2026-02-25
Patch 1

vinodhrajesh:
## Release Notes

* **Chores**
  * Enhanced application deployment with improved container health monitoring, resource management, and automatic restart policies across services
  * Strengthened security by centralizing environment configuration for sensitive credentials
  * Optimized Docker image and application dependency management for improved deployment efficiency

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,4 @@
+POSTGRES_USER=postgres
+POSTGRES_PASSWORD=yoursecurepassword
+POSTGRES_DB=momentum
+NEO4J_AUTH=neo4j/yoursecurepassword
--- a/compose.yaml
+++ b/compose.yaml
@@ -2,16 +2,22 @@ services:
   postgres:
     image: postgres:latest
     container_name: potpie_postgres
-    environment:
-      POSTGRES_USER: postgres
-      POSTGRES_PASSWORD: mysecretpassword
-      POSTGRES_DB: momentum
+    env_file:
+      - .env
     ports:
       - "5432:5432"
     volumes:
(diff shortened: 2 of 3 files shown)


vinodhrajesh:
access restricted to sonarqube

yashkrishan (changes requested):
I would not merge this as-is. The direction is reasonable, but this PR currently breaks local startup and needs a revision before it is safe to merge.

Blocking issues:

1. `compose.yaml` is invalid YAML around the Neo4j service. `env_file` and `environment` are mis-indented, and `docker compose config` fails with `yaml: line 27: did not find expected key`.

2. The Dockerfile switches to `USER appuser`, but `supervisord.conf` still configures both programs with `user=root`. A non-root supervisord process cannot spawn child processes as root, so Celery/Gunicorn are likely to fail at runtime. Please either keep supervisord as root and drop privileges intentionally per process, or update the supervisor program users to the non-root user.

3. The Postgres volume mount changes from `/var/lib/postgresql` to `/var/lib/postgresql/data`. Existing named volumes created with the old layout can appear empty or fail initialization when remounted at the data directory. This needs either a migration note/plan or should avoid changing the mount path in this PR.

4. Committing a real `.env` with changed default credentials makes the documented README/app settings drift from Compose. Please use `.env.example` or update the existing template/docs instead, and keep credentials consistent across the app config.

Please verify the revision with at least `docker compose config` and a clean `docker compose up` before re-requesting review. Also check all the coderabbit comments.

PR #659, opened 2026-02-27
fix: startup-blocking SyntaxError and NameError prevent server from booting

arnavp27:
## Problem

When trying to run the server locally following the README setup, the FastAPI 
worker crashes immediately on boot before serving any requests. The server 
cannot start at all due to 3 bugs hit in sequence during import:

**Bug 1 & 2 — `code_changes_manager.py`**
Two separate `try:` blocks had their body code accidentally de-indented to the 
outer scope, making them syntactically invalid. Python requires every `try:` to 
be followed by `except` or `finally` — but instead it found unindented code, 
causing:
SyntaxError: expected 'except' or 'finally' block

line 4878: if/else block in update_file_lines_tool
line 5362: git_diffs block in show_diff_tool

These look like accidental de-indentation from a merge or refactor.

**Bug 3 — `history_processor.py`**
`ModelResponse` is used as a type annotation on line 494 inside 
`TokenAwareHistoryProcessor._strip_problematic_tool_calls`, but it was never 
imported. Only `ModelMessage` was imported from `pydantic_ai.messages`.
NameError: name 'ModelResponse' is not defined

## Fix

- Re-indented the two misplaced blocks back inside their `try:` bodies in 
  `code_changes_manager.py`
- Added `ModelResponse` to the existing import line in `history_processor.py`:
  `from pydantic_ai.messages import ModelMessage, ModelResponse`

## Verification

After the fixes, the server boots and the health check passes:
$ curl http://localhost:8001/health
{"status":"ok","version":"f1674b80"}

## Notes

Anyone trying to run potpie locally from the current `main` branch will hit 
these errors and be unable to start the server. No logic was changed — 
only indentation and a missing import were fixed.

## Release Notes

**Refactor**
- Internal infrastructure improvements to diff processing and storage with enhanced data persistence.
- Updated internal dependencies to support extended response handling capabilities.

### DIFF
--- a/app/modules/intelligence/agents/chat_agents/history_processor.py
+++ b/app/modules/intelligence/agents/chat_agents/history_processor.py
@@ -37,7 +37,7 @@ from app.modules.intelligence.agents.chat_agents.message_compressor import (
     truncate_tool_result_message,
     validate_and_fix_tool_pairing,
 )
-from pydantic_ai.messages import ModelMessage
+from pydantic_ai.messages import ModelMessage, ModelResponse
 
 logger = logging.getLogger(__name__)
 
--- a/app/modules/intelligence/tools/code_changes_manager.py
+++ b/app/modules/intelligence/tools/code_changes_manager.py
@@ -4875,18 +4875,18 @@ def update_file_lines_tool(input_data: UpdateFileLinesInput) -> str:
             db=db,
         )
 
-        if result.get("success"):
-            context_str = ""
-            if result.get("updated_context"):
-                context_start = result.get("context_start_line", result["start_line"])
-                context_end = result.get("context_end_line", result["end_line"])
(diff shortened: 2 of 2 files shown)


shmbhvi101 (commented):
Thanks for the fix attempt — I verified the current main and the startup issue described here no longer reproduces in today’s codebase.
* `code_changes_manager.py` has been refactored into the `code_changes_manager/ package`, so this PR targets outdated paths.
* `history_processor.py` on current main does not use `ModelResponse` in the way described.
* Current files compile cleanly.

**Verification**
The server boots and the health check passes:
$ curl http://localhost:8001/health
{"status":"ok","version":"0c5c7ec0"}

Given that, this PR is stale and conflicts with main,  we will be closing it.

If you still see a startup failure on latest main, please open a fresh PR with:
1. exact repro on latest commit,
2. minimal patch against current file layout.

PR #666, opened 2026-02-28
fix: move PROJECT_PATH from os.getenv() to ConfigProvider

### LINKED ISSUE #638: Move PROJECT_PATH from os.getenv() to ConfigProvider
### Component/Module
Intelligence (Agents/Tools/Prompts)

### Feature Type
Enhancement to Existing Feature

### Problem Statement
`AgentsService.__init__` reads `PROJECT_PATH` directly via `os.getenv()`, bypassing the `ConfigProvider` abstraction used everywhere else in the project. This introduces several issues:

1. **Inconsistent config access**: `PROJECT_PATH` is the only environment variable in the module read directly via `os.getenv()`, while all others go through `ConfigProvider`.
2. **No startup validation**: The value is not validated when the application starts, meaning misconfiguration is only caught at runtime.
3. **Invisible to central config management**: Because it bypasses `ConfigProvider`, the variable is not visible or manageable through the central config layer.

### Proposed Solution
1. Add a `get_project_path()` static method to `config_provider.py` following the same pattern as existing static methods in that file.
2. Replace the direct `os.getenv` call in `agents_service.py` with `ConfigProvider.get_project_path()`.

**Current Implementation:**
- `agents_service.py` line 57 reads: `os.getenv('PROJECT_PATH', 'projects/')` directly.
- `config_provider.py` already has a consistent pattern for all other env vars using `@staticmethod` methods.

**After this change:**
- `PROJECT_PATH` is read through `ConfigProvider` like every other env var.
- `config_provider.py` becomes the single source of truth for all config.

### Use Case
Any developer deploying Potpie should have `PROJECT_PATH` validated and resolved consistently with all other environment variables — through `ConfigProvider` — so misconfigurations are caught early at startup rather than silently failing later.

### Alternatives Considered
Adding inline validation directly in `agents_service.py` but rejected because it still bypasses `ConfigProvider` and duplicates logic that already belongs in the config layer.

### Additional Context
**Benefits:**
- Consistent config access across the entire codebase
- Startup validation through the central config layer
- Follows the 12-factor app config principle: https://12factor.net/config

**Where to look:**
- `agents_service.py` — line 57
- `config_provider.py` — see existing `@staticmethod` methods for the pattern to follow

**Skill level needed:** Basic Python. No prior knowledge of this codebase required, the pattern to follow is already there in `config_provider.py`.

Feel free to comment if you have any questions before you start!

arnavp27:
## Summary

Fixes #638

- Added `get_project_path()` static method to `ConfigProvider` in `config_provider.py`, following the existing `@staticmethod` pattern
- Replaced direct `os.getenv("PROJECT_PATH", ...)` calls in `agents_service.py` and `search_service.py` with `ConfigProvider.get_project_path()`
- Removed now-unused `import os` and `from pathlib import Path` from both service files

`search_service.py` had the identical issue (same raw `os.getenv` call) so it was fixed in the same PR.

## Summary by CodeRabbit

* **Refactor**
  * Centralized project path resolution into a single provider used by services.
  * Updated services to obtain the project path from the centralized provider, reducing duplication and simplifying configuration handling across the app.

[Review Change Stack](https://app.coderabbit.ai/change-stack/potpie-ai/potpie/pull/666?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Dsantra92 (dismissed):

Dsantra92:
@arnavp27 can you please resolve the conflicts with the main branch? Otherwise it's good to go.

arnavp27:
@Dsantra92 I have resolved it.

Dsantra92 (changes requested):
LGTM

Dsantra92 (approved):

PR #676, opened 2026-03-03
Fix #675: Stack Trace Not Truncated in Production Logs  #645

### LINKED ISSUE #675: Stack Trace Not Truncated in Production Logs  #645
The production log sink writes the full exception traceback on every error. Issue #645 claimed this was fixed to cap at 10 lines, but that truncation code does not exist in the file — the raw traceback string is passed through as-is.

## Current Behavior

In `app/modules/utils/logger.py` at line 134, the traceback is retrieved and passed directly to `filter_sensitive_data()` with no truncation applied:

```python
exc.get("traceback", "")  # passed directly — no line cap
```

Every exception writes the complete traceback to production logs, including full internal file paths and third-party library frames.

## Expected Behavior

The traceback written to production logs should be capped at roughly the last 10 lines — enough to identify the error location without bloating logs or exposing full internal module paths.

## Why This Matters

- **Log volume**: Deep tracebacks (especially from SQLAlchemy, httpx, or async chains) can run to 60–100 lines per exception and are written on every retry or failure.
- **Information exposure**: Full paths like `/app/modules/...` and third-party library internals appear in logs that may flow to external aggregators.
- **Stale fix claim**: Issue #645 is marked resolved but the fix was never applied, creating a false sense of security.

## Suggested Fix Direction

The change is small and localized to `logger.py:134`:

1. Split the traceback string on `\n`
2. Take the last N lines (e.g. `[-10:]`) to preserve the most relevant frames
3. Rejoin and pass the result to `filter_sensitive_data()`

A few judgment calls to make as part of this fix:
- **How many lines?** 10 is a reasonable starting point; adjust based on what your log aggregator shows.
- **Head or tail?** Taking the tail (`[-10:]`) gives the immediate failure frames, which is usually most useful. Taking the head gives the origin call — consider what's more actionable in your setup.

The fix is a one-liner at the truncation site; no changes needed elsewhere.

## Notes for Contributors

- **Where to look**: `app/modules/utils/logger.py` — the `production_log_sink` function, line 134.
- **Goal**: cap traceback length before it is written to logs; no change to log format or filtering behavior.
- **Scope**: entirely within `logger.py` — no other files need to change.

itniuma2026:
## Summary

Truncate exception tracebacks in the production log sink to the last 10 lines before passing to filter_sensitive_data(), preventing log bloat and reducing exposure of internal file paths.

## Approach

In the `production_log_sink` function in `app/modules/utils/logger.py`, locate line 134 where `exc.get('traceback', '')` is retrieved. Split the traceback string on newlines, take the last 10 lines using `[-10:]`, rejoin them, and pass the result to `filter_sensitive_data()`. Specifically, replace the direct retrieval with something like: `tb = exc.get('traceback', ''); tb_truncated = '\n'.join(tb.splitlines()[-10:]) if tb else ''` and then use `tb_truncated` in place of the raw traceback string.

## Files Changed

- `app/modules/utils/logger.py`

## Related Issue

Fixes #675

## Summary by CodeRabbit

* **Refactor**
  * Simplified logging system architecture and configuration.
  * Exception tracebacks now display only the last 10 lines for improved readability instead of full stack traces.


itniuma2026:
Closing this PR due to unintended scope — the diff incorrectly removes 251 lines of unrelated production logging infrastructure (log_data construction, stdout sink, InterceptHandler, configure_logging). Only the traceback truncation change in lines 129-134 was intended. Will resubmit with a targeted fix.

PR #678, opened 2026-03-05
Add SSE parsing status endpoint with Redis event-driven updates

prajwalsuryawanshi:
This PR adds Server-Sent Events support for parsing status so clients can subscribe to live project status changes instead of relying on frontend polling.

**What changed**
- Added SSE mode to parsing status routes:

>   - app\api\router.py
>   - app\modules\parsing\graph_construction\parsing_router.py
> 

- Added parsing status stream generator in:
  

> - app\modules\parsing\graph_construction\parsing_controller.py

- Added Redis status event publishing on project status updates in:

>   - app\modules\projects\projects_service.py

**Behavior**
- Existing polling behavior is preserved by default.
- SSE is enabled via query param `stream=true` on the parsing status endpoint.
- endpoint -` /parsing-status/{project_id}`
- Stream emits:
  - status (state update)
  - heartbeat (keep-alive)
  - complete (terminal state)
  - error (stream/access/timeout failure)
- Access and project visibility checks are still enforced before streaming.
- Stream auto-closes on terminal states (ready/error), disconnect, or timeout.

**Why this change**
- Reduces unnecessary repeated DB polling from clients.
- Moves parsing status updates toward event-driven delivery using Redis stream events.
- Keeps migration safe by maintaining backward compatibility for existing clients.

**Notes**
- Frontend EventSource migration is not included in this PR (backend support only).

## Summary by CodeRabbit

* **New Features**
  * Optional real-time parsing status streaming (SSE) with heartbeats, automatic completion, and fallback to existing non-streaming responses.
  * Project status updates now publish live events so subscribers receive status changes immediately.

* **Improved**
  * More robust error handling, clearer terminal/completion signaling, and safer background-task tracking during parsing flows.


prajwalsuryawanshi (commented):
Fixed and improved stream robustness

prajwalsuryawanshi:
@shmbhvi101 @nndn @dhirenmathur  please have a look on this one.

yashkrishan (changes requested):
Thanks for adding parsing-status SSE. The direction is useful, but I would not merge this yet without tightening the runtime behavior and adding tests.

Concerns to address before merge:

1. `stream_parsing_status` creates a new synchronous Redis client per SSE request and blocks `xread` inside `asyncio.to_thread`. This can consume one Redis connection plus one worker thread per connected client. Please either reuse/align with the existing Redis stream manager pattern in the app or add clear bounds/backpressure so many open parsing-status streams do not exhaust the server.

2. `ProjectService.update_project()` now performs a synchronous Redis publish after every committed status update. Even though failures are swallowed, Redis latency is now on the parsing status update path. Please consider moving this behind a small publisher abstraction/shared manager or otherwise ensure status updates are not delayed by Redis under load.

3. The SSE stream only advances when project status changes go through `ProjectService.update_project(..., status=...)`. Any direct mutation of `Project.status` will leave clients on heartbeats until timeout even though the DB changed. Please either centralize all status writes through the publisher path or make the stream periodically reconcile with the DB.

4. The same `stream=true` route logic is duplicated in both `app/api/router.py` and `app/modules/parsing/graph_construction/parsing_router.py`. Please extract a small helper or keep one source of truth so future fixes do not drift.

5. This needs tests. At minimum, please cover the SSE initial event, a Redis-triggered status update, terminal `ready`/`error` completion, Redis publish on `update_project`, and disconnect/error handling. Streaming behavior is very easy to regress without automated coverage.

Once those are addressed, please verify with an end-to-end parse flow that a connected `/parsing-status/{project_id}?stream=true` client receives the expected `submitted -> parsed -> inferring -> ready/error` progression.

PR #721, opened 2026-04-02
fix: JSON content-type in auth routes + traceback truncation (#634, #675)

### LINKED ISSUE #634: [Bug] Signup endpoint returns JSON body with text/plain content-type header

### LINKED ISSUE #675: Stack Trace Not Truncated in Production Logs  #645

omnialp:
### #634: Signup endpoint returns JSON body with text/plain content-type header
- Replaced 17 instances of `Response(content=json.dumps(...))` with `JSONResponse(content=...)` in `app/modules/auth/auth_router.py`
- All auth endpoints now correctly return `Content-Type: application/json`
- Consistent with login/SSO endpoints that already used `JSONResponse`

### #675: Stack Trace Not Truncated in Production Logs
- Truncated traceback output to last 10 lines in `app/modules/utils/logger.py`
- Reduces log volume from 60-100 line tracebacks per exception
- Limits exposure of internal file paths and third-party library frames

## Changes
- `app/modules/auth/auth_router.py`: 17 Response→JSONResponse replacements
- `app/modules/utils/logger.py`: traceback truncation to 10 lines

Fixes #634, Fixes #675

## Summary by CodeRabbit

* **Bug Fixes**
  * Signup and custom-token endpoints now return proper JSON payloads (native objects), fixing inconsistent/serialized responses and ensuring clients receive consistent, parseable responses across success and error cases.
  * Adjusted a GitHub-related signup error payload to correct its structure for clearer client handling.

* **Refactor**
  * Error logging now truncates tracebacks to the last 10 lines, reducing log noise while preserving relevant context.

### DIFF
--- a/app/modules/auth/auth_router.py
+++ b/app/modules/auth/auth_router.py
@@ -139,12 +139,12 @@ class AuthAPI:
 
         # Validate
         if not uid:
-            return Response(
-                content=json.dumps({"error": "Missing uid"}), status_code=400
+            return JSONResponse(
+                content={"error": "Missing uid"}, status_code=400
             )
         if not email:
-            return Response(
-                content=json.dumps({"error": "Missing email"}), status_code=400
+            return JSONResponse(
+                content={"error": "Missing email"}, status_code=400
             )
 
         async_user_service = AsyncUserService(async_db)
@@ -175,10 +175,10 @@ class AuthAPI:
 
             if not user:
                 logger.error(f"SSO user {link_to_user_id} not found in database!")
-                return Response(
-                    content=json.dumps(
+                return JSONResponse(
+                    content=
                         {"error": "User not found. Please sign in again."}
-                    ),
+                    ,
                     status_code=404,
                 )
 
@@ -196,8 +196,8 @@ class AuthAPI:
 
             if existing_github:
                 logger.info(f"GitHub already linked to user {user.uid}")
-                return Response(
-                    content=json.dumps(
+                return JSONResponse(
+                    content=
                         _signup_response_with_custom_token(
                             {
                                 "uid": user.uid,
@@ -205,7 +205,7 @@ class AuthAPI:
                                 "needs_github_linking": False,
                             }
                         )
-                    ),
+                    ,
                     status_code=200,
                 )
 
@@ -231,13 +231,13 @@ class AuthAPI:
                     f"GitHub account {provider_uid} is already linked to user {existing_provider_with_uid.user_id}, "
                     f"cannot link to user {user.uid}"
                 )
-                return Response(
-                    content=json.dumps(
+                return JSONResponse(
+                    content=
                         {
                             "error": error_message,
                             "details": f"GitHub account {provider_uid} is already linked to user {existing_provider_with_uid.user_id}, cannot link to user {user.uid}",
                         }
-                    ),
+                    ,
                     status_code=409,  # Conflict
                 )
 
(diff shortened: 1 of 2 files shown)


omnialp:
Thanks @coderabbitai! Addressed all feedback:

- ✅ Fixed formatting on multi-line JSONResponse blocks (cleaned up indentation)
- ✅ Removed unused  import
- ✅ Verified syntax passes py_compile

The  patterns in f-strings are intentional per project convention — the other auth routes in the same file use the same pattern.

Dsantra92:
@omnialp can you fix the merge conflicts and the coderabbit issue? this looks good to merge.

PR #731, opened 2026-04-08
fix: remove duplicate create_user call in setup_dummy_user (#726)

### LINKED ISSUE #726: [Bug] High: create_user called twice in setup_dummy_user — causes IntegrityError or corrupts uid
## Summary

In `user_service.py`, `setup_dummy_user` calls `create_user` twice with the same user object. The second call either raises an `IntegrityError` or silently overwrites `uid` with an empty string, corrupting the dummy user setup flow.

## File

`app/modules/users/user_service.py`, `setup_dummy_user` method (~lines 129-130)

## Description

```python
uid, message, error = user_service.create_user(user)  # Creates user correctly
uid, _, _ = user_service.create_user(user)             # Bug: duplicate call!
```

The second `create_user(user)` call is identical to the first. Since the user was already created, it will either:
- Raise a database `IntegrityError` (unique constraint violation), rolling back and logging an error
- Return an empty/error `uid`, silently overwriting the correct value from the first call

## Impact

- Development mode initialization (`setup_dummy_user`) can fail entirely
- If it doesn't fail, `uid` is silently set to an incorrect value
- Any downstream logic relying on `uid` from this function will behave incorrectly

## Suggested Fix

Remove the duplicate call entirely:

```python
# Before (broken):
uid, message, error = user_service.create_user(user)
uid, _, _ = user_service.create_user(user)  # remove this line

# After (fixed):
uid, message, error = user_service.create_user(user)
```

sAchin-680:
### fix: remove duplicate `[create_user]` call in `[setup_dummy_user]`

`[setup_dummy_user]` was calling [create_user] twice with the same user object. The second call would either raise an `IntegrityError` (unique constraint) or silently overwrite [uid] with an empty string — either way breaking dev-mode initialization.

### What changed

- Removed the duplicate [create_user]call in `user_service.py`
- Moved the log line inside the else block so it only fires when a user is actually created

### Tests
Added two unit tests to [TestSetupDummyUser]:
  - Asserts `[create_user]` is called exactly once for a new user
  - Asserts `[create_user]` is skipped entirely when the user already exists

Fixes #726 

## Release Notes

* **Bug Fixes**
  * Resolved duplicate user creation that occurred during system initialization, ensuring users are created only once.

* **Tests**
  * Added comprehensive test coverage for user setup functionality.

### DIFF
--- a/app/modules/users/user_service.py
+++ b/app/modules/users/user_service.py
@@ -110,9 +110,7 @@ class UserService:
                 provider_username="self",
             )
             uid, message, error = user_service.create_user(user)
-
-        uid, _, _ = user_service.create_user(user)
-        logger.info(f"Created dummy user with uid: {uid}")
+            logger.info(f"Created dummy user with uid: {uid}")
 
     def get_user_by_uid(self, uid: str):
         try:
--- a/tests/unit/users/test_user_service.py
+++ b/tests/unit/users/test_user_service.py
@@ -109,3 +109,56 @@ class TestUserServiceUpdateLastLogin:
         message, error = user_service.update_last_login("nonexistent", "token")
         assert error is True
         assert "not found" in message.lower()
+
+
+class TestSetupDummyUser:
+    def test_create_user_called_once_when_user_does_not_exist(self, mock_db, monkeypatch):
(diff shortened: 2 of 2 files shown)


akothari21:
@Dsantra92 Please review this

PR #738, opened 2026-04-17
feat: add dead code detection agent

Suhel-Kap:
What this does

  Adds a new system agent, dead_code_agent, that finds unreachable functions, classes, and interfaces by querying the knowledge graph directly rather than doing text-based grep searches.

  It introduces one new Neo4j tool (find_unreferenced_nodes) that returns all FUNCTION, CLASS, and INTERFACE nodes with zero incoming edges. The agent uses that as a starting point, then filters out legitimate false positives: entry points, framework decorators (@app.route, @celery.*, etc.), test helpers, and any code that might be reached through dynamic dispatch.

  Files changed

  - app/modules/intelligence/tools/code_query_tools/find_unreferenced_nodes_tool.py — new Neo4j tool
  - app/modules/intelligence/agents/chat_agents/system_agents/dead_code_agent.py — new agent following the same pattern as blast_radius_agent.py
  - app/modules/intelligence/tools/tool_service.py — registers find_unreferenced_nodes
  - app/modules/intelligence/agents/agents_service.py — registers dead_code_agent

  Why graph-native

  Text-based dead code detection (grep, regex) produces a lot of noise — it misses aliased calls, dynamic dispatch, and cross-file references. Querying Neo4j for nodes with no
  incoming edges is more reliable because the relationships in the graph already encode actual call structure, not just name proximity.

  Testing

  The agent can be exercised through the existing conversation API by selecting dead_code_agent as the agent ID. I don't have a live Neo4j instance to run integration tests against,
  but the tool follows the same structure as get_node_neighbours_from_node_id_tool.py which has existing test coverage.

  Happy to add unit tests for the Cypher query logic if that's useful.

## Release Notes

* **New Features**
  * Dead Code Detection Agent: A new analysis tool that identifies unreferenced and unused code throughout your codebase, helping maintain code quality and reduce technical debt by detecting code that is no longer being referenced.

### DIFF
--- a/app/modules/intelligence/agents/agents_service.py
+++ b/app/modules/intelligence/agents/agents_service.py
@@ -30,6 +30,7 @@ from .chat_agent import AgentWithInfo, ChatContext
 from .chat_agents.system_agents import (
     blast_radius_agent,
     code_gen_agent,
+    dead_code_agent,
     debug_agent,
     integration_test_agent,
     low_level_design_agent,
@@ -168,6 +169,14 @@ class AgentsService:
                 name="Specification Generation Agent",
                 description="An agent specialized in generating comprehensive technical specifications from user requests through a systematic 7-step process.",
                 agent=specgen.SpecGenAgent(
+                    llm_provider, tools_provider, prompt_provider
+                ),
+            ),
+            "dead_code_agent": AgentWithInfo(
+                id="dead_code_agent",
+                name="Dead Code Detection Agent",
+                description="Identifies unreachable and unused functions, classes, and interfaces by querying the knowledge graph for nodes with zero incoming edges. Returns a grouped report with file paths, line ranges, and confidence levels.",
+                agent=dead_code_agent.DeadCodeAgent(
                     llm_provider, tools_provider, prompt_provider
                 ),
             ),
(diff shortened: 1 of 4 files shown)


yashkrishan (changes requested):
Thanks for adding the dead-code agent. The feature direction is useful, but I would not merge this yet because the core query can miss real dead code and the prompt/tool contract is inconsistent.

Blocking concerns:

1. `find_unreferenced_nodes` counts any incoming non-`IS_LEAF` relationship as a reference. The graph also contains structural relationships such as `CONTAINS` from files to functions/classes, so genuinely uncalled functions can still have an incoming edge and be excluded from the result. For dead-code detection this should count only semantic usage relationships, e.g. `CALLS` and possibly `REFERENCES`, not all relationship types.

2. The agent prompt says that if the user specifies files, directories, or node types, those filters should be passed to `find_unreferenced_nodes`. The tool schema only supports `project_id`, `node_types`, and `limit`, so file/directory filtering is currently impossible. Please either add the filters to the tool or remove that instruction from the prompt.

3. This needs tests around the Neo4j query semantics. At minimum, please cover a fixture where a function has a structural incoming `CONTAINS` edge but no callers and is still returned; a called function is excluded; `node_types` is respected; and `limit` is applied.

4. Please add a small smoke test that `dead_code_agent` is registered/instantiates with the expected tools. Agent/tool registration issues tend to fail late unless there is coverage.

Once the query semantics and tests are in place, this should be much safer to evaluate.

PR #748, opened 2026-04-22
unbound local error while parsing using priority 3 github token

BrhKmr23:
Repo is assigned later in _clone_to_repo_manager, so Python treats it as local for the whole function. On the env-token path we never hit those assignments, but we still called Repo.clone_from before Repo was set — UnboundLocalError.

Fix: call _get_git_imports() before clone_from, and drop the duplicate import after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a cloning initialization issue that could cause failures in certain environment configurations.
  * Ensures repository cloning and subsequent bare-repo setup use a consistent initialization sequence.
  * Improves reliability of import and initialization steps during clone, reducing intermittent errors.

### DIFF
--- a/app/modules/parsing/graph_construction/parsing_helper.py
+++ b/app/modules/parsing/graph_construction/parsing_helper.py
@@ -1893,12 +1893,8 @@ class ParseHelper:
                 method="environment_token_bare_clone",
             )
 
-            # Import Repo before clone_from: a later `Repo = ...` assignment in this
-            # block would make `Repo` local for the whole function and cause
-            # UnboundLocalError on Repo.clone_from (see Priority 3 bare clone path).
-            _, _, Repo = _get_git_imports()
-
             try:
+                _, _, Repo = _get_git_imports()
                 # Clone as bare repository
                 Repo.clone_from(
                     clone_url,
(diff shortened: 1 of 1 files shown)


app/modules/parsing/graph_construction/parsing_helper.py
```diff
@@ -1893,6 +1893,11 @@ async def _clone_to_repo_manager(
                 method="environment_token_bare_clone",
             )
 
+            # Import Repo before clone_from: a later `Repo = ...` assignment in this
+            # block would make `Repo` local for the whole function and cause
+            # UnboundLocalError on Repo.clone_from (see Priority 3 bare clone path).
```
Dsantra92: No need for this comment and put into the try block?

Dsantra92 (changes requested):

app/modules/parsing/graph_construction/parsing_helper.py
BrhKmr23: Resolved in latest commit

Dsantra92 (approved):
LGTM

PR #762, opened 2026-05-04
Consolidate repo_name and repo_path into a single repository identifier

### LINKED ISSUE #517: Consolidate repo_name and repo_path into a Single Repository Identifier
Currently, the codebase accepts `repo_name` and `repo_path` as separate variables throughout the parsing and project creation flows. This introduces several issues:

1. **Dual variable management**: Code must handle both `repo_name` and `repo_path` separately, requiring conditional logic and validation to ensure exactly one is provided.
2. **Scattered resolution logic**: Repository path resolution, cloning, and normalization logic is distributed across multiple layers (controller, service, helper).
3. **Late-stage detection**: The system only determines whether a repository is local vs. remote during the controller phase, after validation.
4. **Redundant conversions**: Current code performs redundant conversions between `repo_path` and `repo_name` (e.g., extracting name from path).

## Current Implementation

- **ParsingRequest schema** requires exactly one of `repo_name` or `repo_path` (validation in `parsing_schema.py`).
- **Auto-detection logic** in `parsing_controller.py` checks if `repo_name` is actually a filesystem path.
- **Path extraction** scattered across multiple files when needed.
- **RepoDetails model** contains both fields but with optional `repo_path`.

## Proposed Solution

1. **Single unified identifier**: Replace dual variables with a single `repository` identifier that can represent either:
   - A remote GitHub repository (`owner/repo`)
   - A local filesystem path (`/path/to/repo`)

2. **Early resolution**: Move repository resolution logic to the earliest possible point (controller entry or new dedicated resolver service)

3. **Normalize and clone early**: 
   - Detect repository type (local vs. remote) immediately
   - Clone remote repositories to a local path if needed
   - Validate filesystem paths early

4. **Downstream simplification**: All downstream services receive a `RepoDetails` object with:
   - `repository_identifier`: The original identifier (for tracking)
   - `repo_path`: The resolved local filesystem path
   - `repo_name`: Normalized repository name (extracted from path or GitHub identifier)
   - `is_local`: Boolean flag indicating if it's a local repository

## Benefits

- Simpler API contracts (one parameter instead of two)
- Cleaner validation logic
- Centralized repository resolution

haroon0x:
## Changes

  - Added `RepositoryResolver` to classify repository identifiers as local paths
  or remote repos.
  - Added `repository_identifier` to `ParsingRequest`.
  - Kept legacy `repo_name` / `repo_path` support with a deprecation warning.
  - Moved path detection from `parsing_controller.py` into schema-level request
  normalization.
  - Added `is_local` to `RepoDetails`.
  - Updated controller logic so local dev requests keep the normalized
  `repo_name` instead of clearing it.
  - Rebased on latest `main` and kept the newer `ProjectSandbox.ensure()`
  parsing flow.

  ## Notes

  During rebase, `parsing_service.py` conflicted because this branch changed the
  old clone/setup flow, while `main` had replaced that flow with sandbox-based
  parsing. I kept the sandbox flow from `main` and dropped the stale legacy
  conversion because it is no longer used.

  ## Testing

  - `git diff --check`
  - `python3 -m py_compile app/modules/parsing/graph_construction/
  parsing_controller.py app/modules/parsing/graph_construction/parsing_schema.py
  app/modules/parsing/utils/repository_resolver.py`

  I also tried running the targeted parsing tests, but the local environment is
  missing `redis`, so pytest fails while loading `tests/conftest.py`.

  ## Follow-up work

  This does not fully remove downstream `repo_name` / `repo_path` usage yet. To
  fully close #517, we still need to:

  - Add `repository_identifier` to `RepoDetails`.
  - Add `is_local` to `ParsingRequest`.
  - Use `is_local` instead of checking `repo_path`.
  - Remove remaining fallback basename extraction from the controller.
  - Audit project registration and lookup paths that still depend on both
  `repo_name` and `repo_path`.
  - Update API docs and examples to prefer `repository_identifier`.

### DIFF
--- a/app/modules/parsing/graph_construction/parsing_controller.py
+++ b/app/modules/parsing/graph_construction/parsing_controller.py
@@ -57,27 +57,12 @@ class ParsingController:
         parse_helper = ParseHelper(db)
         parsing_service = ParsingService(db, user_id)
 
-        # Auto-detect if repo_name is actually a filesystem path
-        if repo_details.repo_name and not repo_details.repo_path:
-            is_path = (
-                os.path.isabs(repo_details.repo_name)
-                or repo_details.repo_name.startswith(("~", "./", "../"))
-                or os.path.isdir(os.path.expanduser(repo_details.repo_name))
-            )
-            if is_path:
-                # Move from repo_name to repo_path
-                repo_details.repo_path = repo_details.repo_name
-                repo_details.repo_name = repo_details.repo_path.split("/")[-1]
-                logger.info(
-                    f"Auto-detected filesystem path: repo_path={repo_details.repo_path}, repo_name={repo_details.repo_name}"
-                )
+        # Repository identifier resolution (local vs remote) is handled
+        # in ParsingRequest.__init__ via RepositoryResolver. At this point
+        # repo_details.repo_name is always populated, and repo_details.repo_path
+        # is set only for local repositories.
 
-        if config_provider.get_is_development_mode():
-            # In dev mode: if both repo_path and repo_name are provided, prioritize repo_path (local)
-            if repo_details.repo_path and repo_details.repo_name:
-                repo_details.repo_name = None
-            # Otherwise keep whichever one is provided as-is
-        else:
+        if not config_provider.get_is_development_mode():
             # In non-dev mode: if repo_name is None but repo_path exists, extract repo_name from repo_path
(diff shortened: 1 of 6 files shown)


Dsantra92:
The original linked issue is no longer required under our current roadmap. Sorry for the inconvenience.

PR #749, opened 2026-04-22
Feat/context engine

nndn:

PR #765, opened 2026-05-06
fix(tests): repair CI failures after context-engine merge

nndn:
## Summary
Two CI failures introduced by the recent context-engine merge (#749), plus the SonarQube findings the PR surfaced on the freshly-added test file.

### Test fixes
- **`tests/integration-tests/parsing/test_parsing_service.py`** — drop the legacy `TestAnalyzeDirectory` class (4 tests) and the obsolete `test_inference_service_failure`. `ParsingService.analyze_directory` was removed in #749 and replaced with `analyze_workspace` (different signature: `WorkspaceHandle` instead of `extracted_dir` / `language`). Equivalent coverage already lives in `tests/unit/parsing/test_sandbox_parse_flow.py` (`test_analyze_workspace_happy_path`, `_rejects_repo_with_no_parseable_code`, `_propagates_parser_failure`).
- **`tests/integration-tests/sandbox_e2e/test_local_sandbox.py:416`** — loosen the worktree-collision regex to match both legacy (`already checked out`) and modern (`already used by worktree`) git phrasings, so the assertion passes regardless of CI's git version.

### SonarQube fixes (same file)
SonarQube treated the entire newly-added `test_local_sandbox.py` as "new code" against this PR's baseline and surfaced four findings on it. All addressed:
- **S7487 ×2 (MAJOR, reliability)** — synchronous `subprocess.run` inside `async def`. Wrapped the `git rev-parse` call in `asyncio.to_thread`, and replaced `subprocess.run(["ls", …])` with `os.listdir` (cleaner, no shell-out).
- **S7498 (MINOR, maintainability)** — `dict(...)` constructor → literal `{...}` in the kwargs fixture.
- **S5443 (security hotspot)** — hardcoded `/tmp/x` in the absolute-path-rejection test → `/etc/x`. The test only cares that an absolute path is refused.

## Test plan
- [ ] Integration phase passes (was 5 failures: 4 in `TestAnalyzeDirectory` + 1 worktree-regex).
- [ ] SonarQube quality gate flips to passing on the PR (new reliability/maintainability ratings back to A; hotspot cleared).
- [ ] `tests/unit/parsing/test_sandbox_parse_flow.py` continues to cover `analyze_workspace`.

## Summary by CodeRabbit

* **Tests**
  * Refactored integration tests for clearer focus and organization; removed obsolete test cases.
  * Converted a blocking operation to be async-safe to stabilize test execution.
  * Broadened error expectations to accept multiple message variants for robustness.
  * Strengthened path-safety checks by validating rejection of an absolute system path.

### DIFF
--- a/tests/integration-tests/parsing/test_parsing_service.py
+++ b/tests/integration-tests/parsing/test_parsing_service.py
@@ -1,5 +1,5 @@
 """
-Integration tests for ParsingService (parse_directory, analyze_directory) with mocks.
+Integration tests for ParsingService.parse_directory with mocks.
 
 Heavy mocking of ParseHelper, CodeGraphService, InferenceService, ProjectService;
 no real Neo4j/Git/RepoManager required.
@@ -11,7 +11,6 @@ import pytest
 from neo4j.exceptions import ServiceUnavailable
 
 from app.modules.parsing.graph_construction.parsing_helper import (
-    ParsingFailedError,
     ParsingServiceError,
 )
 from app.modules.parsing.graph_construction.parsing_schema import ParsingRequest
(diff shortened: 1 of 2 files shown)


nndn:
Pushed `7918f2c7` to address the SonarQube quality-gate findings on this PR. Reasoning per finding (all in `tests/integration-tests/sandbox_e2e/test_local_sandbox.py`):

**Line 175 — S7487 (sync subprocess in async)**
```diff
- head = subprocess.run(
+ head = await asyncio.to_thread(
+     subprocess.run,
      ["git", "-C", str(worktree), "rev-parse", "--abbrev-ref", "HEAD"],
      capture_output=True, text=True, check=True,
  )
```
We still need a real `git rev-parse` to assert the worktree's branch matches `handle.branch`, so swapping to `asyncio.create_subprocess_exec` would just churn the call site. `asyncio.to_thread` keeps the call shape and parses-as-CompletedProcess, but takes it off the event loop.

**Line 389 — S7487 (sync subprocess in async)**
```diff
- listing = subprocess.run(
-     ["ls", h_b.local_path],
-     capture_output=True, text=True, check=True,
- )
- assert "ONLY_IN_A.txt" not in listing.stdout
+ assert "ONLY_IN_A.txt" not in os.listdir(h_b.local_path)
```
Shelling out to `ls` was never the right move here — the assertion only needs the directory contents. `os.listdir` is what was actually meant; deleting the subprocess clears the rule cleanly.

**Line 190 — S7498 (constructor → literal)**
```diff
- kwargs = dict(
-     user_id="u1", ...
- )
+ kwargs = {
+     "user_id": "u1", ...
+ }
```
Pure stylistic; the call sites (`**kwargs`) are unaffected.

**Line 508 — S5443 (publicly writable dir)**
```diff
- await local_client.write_file(handle, "/tmp/x", b"x")
+ await local_client.write_file(handle, "/etc/x", b"x")
```
The test asserts that writes to *any* absolute path are refused via `InvalidWorkspacePath`. `/tmp` was incidental, and SonarQube flags hardcoded `/tmp` as a publicly-writable-dir hotspot. `/etc/x` keeps the same semantics (absolute path → must reject) without tripping the rule.

> Worth noting: every line SonarQube flagged was authored by #749 — the PR's own diff is just the regex on line 416. They were attributed to this PR because SonarQube's "new code" baseline for the branch predates the #749 merge. Folding the cleanups in here rather than leaving them dangling on main.

Dsantra92 (approved):
LGTM

PR #232, opened 2025-01-26
feat: Implement incremental update system with snapshot restore 

onyedikachi-david:
Fixes: #221
/claim #221

Issue #221, opened 2025-01-09
Incremental Knowledge Graph updates


dhirenmathur:
# Implement Incremental Knowledge Graph Updates

## Problem
Currently, the knowledge graph system performs a complete rebuild of both code mapping and inference generation on any update, regardless of the change magnitude. This approach is computationally expensive and time-consuming, impacting system performance and resource utilization.

## Proposed Solution
Develop an incremental update system for the knowledge graph that processes only the changed components rather than rebuilding the entire graph.

### Core Functionality
- Implement differential analysis to detect and isolate changes between versions
- Create an incremental update mechanism that preserves existing graph structure
- Develop a system to track and maintain relationship integrity during updates

### Key Features
1. **Change Detection**
   - Identify modified, added, and deleted components
   - Determine affected relationships and dependencies
   - Map impact scope of changes

2. **Relationship Management**
   - Maintain accurate relationship tracking across updates
   - Handle cascading relationship updates
   - Preserve existing relationship metadata

3. **Inference Handling**
   - Generate inferences only for new/modified components
   - Preserve existing valid inferences
   - Update affected inference chains without full regeneration

4. **Performance Optimization**
   - Eliminate full graph reconstruction
   - Minimize parsing operations to affected components only
   - Optimize resource usage during updates

## Technical Considerations
- Must maintain data consistency throughout incremental updates
- Need to implement versioning/tracking for incremental changes
- Requires robust error handling for partial updates
- Should include rollback capability for failed updates

PriyanshGodhani22:
@dhirenmathur is this issue solved otherwise I can submit a plan to it just wanted confirm so that it doesn't replicate the efforts thanks.

chawansrujeeth:
/attempt #221
Hi 
@dhirenmathur 
@yashkrishan 

I am working on this issue and have a simple, incremental approach in mind to avoid rebuilding the Knowledge Graph on every run.

The idea is to persist the extracted graph artifacts (e.g., nx_nodes.jsonl / nx_edges.jsonl), diff the previous snapshot against the newly extracted one using stable identifiers, and apply only the resulting add/delete deltas to Neo4j instead of wiping and rebuilding the graph.

This keeps unchanged parts of the codebase untouched while making ingestion incremental and safer. I’ll open a PR with this approach .

Thanks.

zhaog100:
Hi! I'd like to work on this bounty. I'll fork and submit a PR. Let me know if there are any specific requirements.

[...]

PR #768, opened 2026-05-06
feat(parsing): incremental knowledge-graph updates (#221)

brone1323:
/claim #221

Re-attempts issue #221 per maintainer's [reopen invitation](https://github.com/potpie-ai/potpie/pull/232#issuecomment-2706192005) on the previously-stale PR #232.

## Approach

Different from PR #232 (which used a snapshot/restore mechanism) â€” this PR runs a direct file-level diff against the persisted graph and applies only the changed parts. The full re-parse still happens inside the sandbox runner (we don't touch `parsing_rs`), but the expensive Neo4j cleanup-and-reinsert + Qdrant rebuild + inference docstring regeneration are scoped to the dirty files.

## Pipeline

1. **Hash per file** from parser artifacts. Each file's fingerprint is a SHA-1 over its sorted `(id, type, line, end_line, name, text)` node-tuples plus the edges originating in it. Hashes are persisted as a `content_hash` attribute on the FILE nodes in Neo4j â€” no new schema/migration. The existing full-rebuild path now also seeds these hashes, so the next run on the same project is automatically incremental.

2. **Diff** new vs persisted hashes into `added` / `modified` / `deleted` / `unchanged` buckets.

3. **Apply delta**:
   - `DETACH DELETE` every node whose `file_path` is in the deleted-or-modified set. `DETACH DELETE` drops all attached edges in one hop, so we don't need a separate edge-cleanup pass.
   - Insert the artifacts' nodes whose file is in the added-or-modified set, using the same label assembly the full path uses.
   - `MERGE` every edge whose source or target file is dirty â€” this is the "relationship integrity" requirement from the issue. Cross-file references where one side moved get re-resolved against the freshly-inserted target nodes. Edges between two unchanged files are left alone (they were never deleted).

4. **Update Qdrant** in lock-step: drop points by `file_path` for removed files; index the dirty files' new nodes into the existing collection (no `recreate_collection`).

5. **Skip inference entirely** when no file-level changes are detected.

## Wiring

- `ParsingRequest` gains `full_rebuild: bool = False`. Default route is incremental; callers can opt back into the legacy clean-and-rebuild via `full_rebuild=True`.
- `ParsingService.parse_directory` skips `cleanup_graph` unless `full_rebuild=True`. The incremental path performs surgical deletes per dirty file instead of wiping the whole project.
- First incremental run on a project that has no `content_hash` data falls back to a full rebuild and seeds hashes â€” no migration step required.

## Tests

`tests/unit/parsing/test_incremental_update.py` â€” 14 cases:

- Per-file hashing is stable across runs and bucket-scoped (modifying one file's body or one of its outbound edges only moves that file's hash).
- Diff buckets: added / modified / deleted / unchanged.
- First-run fallback to full rebuild + hash seeding.
- No-op when nothing changed (no DELETE, no INSERT, no inference).
- Modified file scopes the DETACH DELETE to that file's `file_path` only.
- Added file inserts without touching existing nodes.
- Deleted file removes its nodes with no re-insert.
- Cross-file edge `A.foo â†’ B.bar` is re-emitted when `B.py` changes (issue #221's relationship-integrity requirement). CONTAINS edges between two unchanged files are NOT re-emitted (no double-write).
- `affected_node_ids_for_inference` returns only dirty non-FILE nodes â€” used by the inference layer when scoping docstring regeneration.
- `reconstruct_subgraph` keeps only dirty-file nodes with full attributes; cross-file inbound edges survive.

The two existing integration tests in `test_parsing_service.py` that asserted on `cleanup_graph` behaviour are updated to set `full_rebuild=True`, since the incremental default doesn't invoke that path.

## Out of scope

- The in-sandbox parser still parses the whole repo each run; making the parser itself incremental is a much larger lift inside `parsing_rs` / `parser_runner` and the issue body's "minimize parsing operations to affected components only" goal is met for the host-side pipeline (which is where the wall-clock goes).
- Rollback / partial-failure recovery is the same as the existing path: a failed Neo4j write leaves the graph in whatever partial state Neo4j accepts, and the next parse with `full_rebuild=True` recovers cleanly. The hashes are written *after* the delta succeeds, so a crashed run won't fool the next diff into thinking the changes already landed.

## Constraints respected

- No `.github/workflows/*` changes.
- No new feature flags beyond the documented `full_rebuild` opt-out.
- No backwards-compat shims â€” this is the new default and the legacy path is one boolean away.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Incremental graph update pipeline for file-scoped sync and an explicit full-rebuild option.

* **Improvements**
  * Deterministic per-file content hashing to detect changed files and avoid unnecessary work; inference is skipped when no changes.
  * Targeted vector index deletions and optional BM25 vocabulary preservation for safer index updates; vector sync is best-effort.

* **Tests**
  * New and updated tests covering incremental updates, hashing, delta application, and inference behavior.

### DIFF
--- a/app/modules/parsing/graph_construction/code_graph_service.py
+++ b/app/modules/parsing/graph_construction/code_graph_service.py
@@ -95,6 +95,46 @@ class CodeGraphService:
             relationship_count=relationship_count,
         )
         self._store_graph(nx_graph, project_id, user_id)
+        # Seed per-file content hashes so the *next* parse can run
+        # incrementally. Done after the full rebuild succeeds — if the
+        # rebuild fails we don't want stale hashes pointing at a
+        # half-written graph.
+        self._write_file_content_hashes(artifacts, project_id)
+
+    def _write_file_content_hashes(self, artifacts, project_id: str) -> None:
+        """Stamp ``content_hash`` on the FILE nodes we just inserted.
+
+        Invoked from the full-rebuild path so the incremental path has a
+        baseline to diff against on the next run. Imported here rather
+        than at module top to keep the incremental layer optional —
+        if the import fails for any reason the full rebuild itself is
+        unaffected.
+        """
+        try:
+            from app.modules.parsing.graph_construction.incremental_graph_service import (  # noqa: E501
+                compute_file_hashes,
+            )
+        except ImportError:
+            logger.warning(
+                "[GRAPH GENERATION] incremental_graph_service unavailable; "
+                "skipping content_hash seeding",
+                project_id=project_id,
+            )
+            return
+        hashes = compute_file_hashes(artifacts)
+        if not hashes:
+            return
+        rows = [{"file_path": fp, "content_hash": h} for fp, h in hashes.items()]
+        with self.driver.session() as session:
+            session.run(
+                """
+                UNWIND $rows AS row
+                MATCH (n:FILE {repoId: $project_id, file_path: row.file_path})
+                SET n.content_hash = row.content_hash
+                """,
+                project_id=project_id,
+                rows=rows,
+            )
 
     def _store_graph(
         self,
--- a/app/modules/parsing/graph_construction/incremental_graph_service.py
+++ b/app/modules/parsing/graph_construction/incremental_graph_service.py
@@ -220,11 +220,26 @@ class IncrementalGraphService:
         existing_hashes = self._read_existing_file_hashes(project_id)
         new_hashes = compute_file_hashes(artifacts)
 
-        if not existing_hashes:
+        # Fall back to a full rebuild when hashes are absent *or* when
+        # coverage is partial (some FILE nodes lack content_hash). A
+        # partial-seeding gap means files without hashes are treated as
+        # "added", so their old nodes are never deleted and
+        # apoc.create.node() will duplicate them on the next incremental
+        # run. Comparing against the total FILE count in the graph lets
+        # us detect this case and force a clean slate.
+        total_file_count = self._count_file_nodes(project_id)
+        is_unseeded = (
+            not existing_hashes
+            or len(existing_hashes) != total_file_count
+        )
+        if is_unseeded:
             logger.info(
-                "[INCREMENTAL] No existing file hashes for project %s — "
-                "running full rebuild and seeding hashes",
+                "[INCREMENTAL] Incomplete hash coverage for project %s "
+                "(hashed=%d, total_files=%d) — running full rebuild and "
+                "seeding hashes",
                 project_id,
+                len(existing_hashes),
+                total_file_count,
             )
             self.graph_service.store_graph_from_artifacts(
                 artifacts, project_id, user_id
@@ -247,16 +262,27 @@ class IncrementalGraphService:
             return delta
 
         start = time.time()
-        self._apply_delta(
+        qdrant_ok = self._apply_delta(
             artifacts=artifacts,
             project_id=project_id,
             user_id=user_id,
             delta=delta,
         )
-        # Persist hashes for the next run. We update *all* files'
-        # hashes — including unchanged ones — so a corrupted hash on
-        # any file (e.g. from an aborted previous run) self-heals.
-        self._write_file_hashes(project_id, new_hashes)
+        if qdrant_ok:
+            # Persist hashes for the next run. We update *all* files'
+            # hashes — including unchanged ones — so a corrupted hash on
+            # any file (e.g. from an aborted previous run) self-heals.
+            self._write_file_hashes(project_id, new_hashes)
+        else:
+            # Qdrant sync failed. Skip writing hashes so the next parse
+            # sees the same delta and retries the vector update. The
+            # Neo4j graph is already consistent — only the search index
+            # is stale until the retry succeeds.
+            logger.warning(
+                "[INCREMENTAL] Skipping hash write for project %s because "
+                "Qdrant sync failed — will retry on next parse",
+                project_id,
+            )
 
         logger.info(
             "[INCREMENTAL] Applied delta in %.2fs",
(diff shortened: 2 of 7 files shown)


brone1323:
## Demo video

[▶ Watch demo (1.9MB MP4)](https://github.com/brone1323/potpie/releases/download/demo-bounty-221-1778105227438/potpie-incremental-kg.mp4)

Generated programmatically (HTML deck → Playwright recordVideo → ffmpeg). Hosted as a release asset on the fork to comply with the Algora claim requirement.

_Sammy / @brone1323_

brone1323:
## CodeRabbit feedback addressed

Thanks for the thorough review. Pushed three fix-up commits (`4368fc4`, `635db36`, `b436666`) addressing all 4 actionable findings:

1. **`full_rebuild=True` could silently skip cleanup** (`parsing_service.py`) — `cleanup_graph` is now coerced to `True` whenever `full_rebuild=True` so the legacy wipe-then-rebuild path always triggers when the caller asks for a full rebuild, even if they passed `cleanup_graph=False`.

2. **`delete_points_by_node_ids` AND-ed `node_id` and `file_path`** (`qdrant_indexing_service.py`) — Switched the Qdrant `Filter` from `must=[...]` to `should=[...]` so the docstring's "by node_id *or* file_path" semantics actually hold when both arguments are supplied.

3. **Incremental sync wrote to stale base collection** (`incremental_graph_service.py`) — Both the delete and the index call now target the active alias (`get_qdrant_collection_alias`) instead of the base collection name, so writes land in whichever collection the alias currently resolves to (important after a staged full-rebuild has swapped the alias and dropped the original base collection).

4. **Incremental BM25 vocabulary remap** (`qdrant_indexing_service.py` + `incremental_graph_service.py`) — Added a `preserve_bm25_vocabulary` flag to `index_nodes_to_qdrant`. The incremental path opts in: instead of replacing the persisted vocabulary with one built from the dirty subset, the function reads the existing vocab, appends only previously unseen tokens at the end (preserving prior token→index slots), and re-saves. This keeps sparse term indices stable across already-indexed points so older BM25 lookups remain correct.

Note on (4): a fully correct fix for IDF drift would also require a global document-frequency tracker per collection (current incremental IDF still uses `N = len(dirty_nodes)`). Vocabulary stability is the load-bearing correctness piece — IDF drift only nudges ranking, not term match — and a per-collection DF index is a larger change worth a follow-up. Happy to scope that separately if you'd like.

_Sammy / @brone1323_

brone1323:
### Fix-up commit 6dbf5c — addresses 2 new review threads (2026-05-06 23:27 UTC round)

**Thread 1 — Partial hash-seeding detection** (comment 3198110109, line 223)

The old guard (if not existing_hashes) only caught the fully-empty case. If some FILE nodes already had content_hash but others did not, those unhashed files were bucketed as *added* instead of *unchanged* — old nodes were never deleted, and poc.create.node() duplicated them on the next incremental run.

Fix: added _count_file_nodes() (one cheap MATCH … RETURN count(n) Cypher) and compare len(existing_hashes) != total_file_count before entering the incremental path. Any coverage gap forces a full rebuild + re-seed.

**Thread 2 — Hash write must not advance after a failed Qdrant sync** (comment 3198110111, line 566)

_update_qdrant() swallowed Qdrant exceptions, but store_graph_from_artifacts_incremental() still wrote content_hash afterward. That made the miss sticky — the next parse saw "no changes" and silently skipped the Qdrant retry.

Fix: _update_qdrant() and _apply_delta() now return ool (True = Qdrant sync OK). _write_file_hashes() is only called when the bool is True. On failure the hashes are left untouched so the next parse observes the same delta and retries automatically. Neo4j is always updated regardless.

Two new unit tests added in 	est_incremental_update.py covering both behaviours.

---
The 4 earlier threads (3197826119, 3197826144, 3197826154, 3197826163 — alias targeting, BM25 vocab, OR-semantics delete, full\_rebuild cleanup) were already addressed and marked ✅ by CodeRabbit.

brone1323:
### Fix-up d53c7a7 â€” addresses round-3 review (CodeRabbit + SonarQube)

**Thread 1 â€” ImportError handler returned True** (coderabbitai comment 3198709736, line 589)

Valid bug. The `except ImportError` branch in `_update_qdrant()` was returning `True`, which told the caller that Qdrant sync succeeded. The caller then wrote `content_hash`, permanently hiding the stale vector state â€” every subsequent parse would see "no changes" and never retry the vector sync.

Fix: changed `return True` to `return False` with an updated comment explaining the intent. This is consistent with every other failure path in `_update_qdrant()` (lines 622 and 672 both return `False`).

**Thread 2 â€” SonarQube Security Hotspot** (1 hotspot, `compute_file_hashes`)

`hashlib.sha1()` at line 168 was flagged. SHA-1 here is a content-change fingerprint (not used for authentication or integrity verification), so the fix is `hashlib.sha1(usedforsecurity=False)` â€” a Python 3.9+ parameter that explicitly opts out of the security-hash contract and suppresses the SonarQube/bandit warning. Added `# noqa: S324` for bandit parity.

brone1323:
## Round-4 fix-up — commit `1d0575c`

Addresses the three CodeRabbit critical/major comments posted at 2026-05-07T04:02Z.

| Thread | Fix |
|---|---|
| **Line 247** — `is_unseeded` fallback skips cleanup | Added `self.graph_service.cleanup_graph(project_id)` call before `store_graph_from_artifacts()` so stale nodes are wiped before the rebuild, preventing `apoc.create.node()` duplicates. |
| **Line 543** — Neo4j retries not idempotent for newly added files | Changed `files_to_delete = delta.removed \| delta.modified` to `delta.removed \| delta.dirty` (`dirty` = `added \| modified`). Added files are now purged in step 1 so a retry after a failed `_update_qdrant()` can't create duplicate nodes. |
| **Line 763** — boundary nodes materialised with empty attrs by `add_edge` | Refactored `reconstruct_subgraph` to collect all artifact nodes into a `node_lookup` dict and call `_add_node()` on both edge endpoints before `graph.add_edge()`, ensuring unchanged-file boundary nodes are added with full metadata rather than as empty stubs. |

brone1323:
## Round-5 fix-up — commit `f6a22db`

Addresses the three items from the post-`1d0575c` CodeRabbit review (2026-05-07T13:16Z).

| Item | Action | Detail |
|---|---|---|
| Lines 146-147 & 482-483 — `getattr(edge, "source_id"/"target_id")` | **Implemented** | Replaced with direct `edge.source_id` / `edge.target_id` at both call sites (`compute_file_hashes` and `_insert_delta_nodes_and_edges`). Ruff B009; fields are typed on the Edge object. |
| Lines 510-516 — `rel_type` interpolated into Cypher without validation | **Implemented** | Added module-level `_REL_TYPE_RE = re.compile(r"^[A-Z][A-Z0-9_]*$")`; edges whose `rel_type` fails the check are skipped with a WARNING log, blocking Cypher injection from malformed parser output. |
| Lines 393-406 — Cypher delete-and-count simplification | **Skipped** | The suggested FOREACH/collect pattern does not straightforwardly return a deletion count in Neo4j 4.x/5.x (mutating clauses after aggregation). The current `_placeholder` shim is functionally correct; refactoring without test coverage risks a silent regression. |

Dsantra92:
As mentioned, this issue has already been taken care of in our current branch. We really appreciate your work and would love to see you pick up a new issue. Closing this PR for now.

PR #814, opened 2026-05-27
chore(context-engine): bind correlation IDs at job seam + quality pass on reconciliation agent (PR 2 of 3)

yashkrishan:
Second of three PRs. Stacked on PR #813 (wire-up).

## Summary

Two-part change:

**Part 1 — correlation seam at the job boundary**

`application/use_cases/context_graph_jobs.py::handle_process_batch` wraps the `process_batch(...)` call in `with log_context(batch_id=batch.id, pot_id=batch.pot_id):`. Every log emitted by `process_batch` and the reconciliation agent it drives now carries those fields as structured metadata. **Single 3-line addition; zero callees changed.** The 17 log calls inside `process_batch.py` and most of the calls in `pydantic_deep_agent.py` benefit automatically.

**Part 2 — manual quality pass on the top-emitter file**

`adapters/outbound/reconciliation/pydantic_deep_agent.py` has 20 log sites — the biggest concentration in context-engine. Cleaned up:

- Dropped redundant `batch %s` / `for batch %s` / `pot=%s` interpolations now that `log_context` covers them.
- Converted message-baked values to structured kwargs where it improves queryability: `record_type`, `timeout_s`, `tool`, `completed_count`, `finish_called`, `usage`, `blocked`, `allowed`, `cap`, `event_id`.
- Left the bare-message exception / debug lines alone — they don't benefit from kwargs and the surgical rule says don't touch them.

**Net:** 11 of the 20 sites in `pydantic_deep_agent.py` converted, plus the one seam binding.

## Why not also do the other 4 top-emitter files manually

Originally PR 2 was scoped to do manual quality on the top 5 emitter files. Mid-PR I noticed that the remaining 4 (`episodic`, `hybrid_graph`, `github/agent_tools`, `process_batch` itself) are simpler patterns that fit cleanly into PR 3's mechanical AST sweep. Doing them by hand here and then re-sweeping with PR 3 would be duplicate work. Cleaner split: PR 2 = highest-judgment file by hand, PR 3 = everything else by AST. That keeps both PRs small and reviewable.

## What this PR does NOT do

- Doesn't touch the other 4 top-emitter files (handled by PR 3).
- Doesn't touch the 48 long-tail files (handled by PR 3).
- Doesn't change anything outside context-engine.

## Test plan

- [x] `uv run python -m compileall -q app/src/context-engine` clean.
- [x] `configure_observability()` still installs exactly one managed handler.
[...]

PR #815, opened 2026-05-27
chore(context-engine): migrate logger sourcing + additive kwargs sweep across all callers (PR 3 of 3)

yashkrishan:
Third and final of the context-engine observability PRs. Stacked on PR #814.

## Summary

Two coordinated changes across 51 files in context-engine.

### Part 1 — logger sourcing migration (50 files)

Replace `logger = logging.getLogger(__name__)` with `logger = get_logger(__name__)`, importing from the observability package alongside the existing `import logging`.

- The router file (`adapters/inbound/http/api/v1/context/router.py`) uses `_logger` / `_audit_logger` names and a dynamic name arg, so it was migrated explicitly outside the bulk sed.
- Tests and benchmarks intentionally skipped — they don't run in production.

### Part 2 — additive kwargs sweep (129 sites across 38 files)

For every `logger.<level>(<msg>, <positional args>...)`, **add** kwargs alongside the positionals so values become queryable fields in logfire. **The message and positional args are unchanged** — zero readability regression.

```python
# Before
logger.exception("github_get_pull_request %s#%s failed", repo_name, pr_number)
# After
logger.exception("github_get_pull_request %s#%s failed", repo_name, pr_number,
                 repo_name=repo_name, pr_number=pr_number)
```

**Field naming:**
- If the message contains `key=%s`, the kwarg name comes from that key (e.g. `"receipt=%s"` → `receipt=...`).
- Otherwise the kwarg uses the variable name from the positional arg. Dotted paths get synthesized (`state.pot_id` → `state_pot_id`).
- Complex expressions (`len(x)`, `x:.2f`, `x!r`, function calls, subscripts) are left untouched — those need per-site judgment on field naming and stay as a long-tail follow-up.
- Kwarg names that collide with existing kwargs (e.g. `exc_info=True`) are skipped, not overwritten.

Each transformed file is re-parsed before write; the script bails per-file if the rewrite would be malformed.

## Why both parts ship together

Plain `logging.getLogger(__name__)` raises `TypeError` on arbitrary kwargs. Migrating to the observability package's `StructuredLogger` is a prerequisite for the kwargs to be accepted and routed into `record.obs_fields` (and on to logfire as fields). Splitting them into separate PRs would leave the codebase momentarily broken between them.

## Why an earlier destructive sweep was reverted mid-PR

The first sweep I tried was *destructive* — replace `logger.X("msg %s", x)` with `logger.X("msg", x=x)`. It worked syntactically but mangled ~10 messages cosmetically (`"github_get_pull_request# failed"`, trailing `:` left dangling, etc.) because the script tried to strip `%s` placeholders without knowing where the natural sentence boundaries were. Reverted and switched to the additive approach (same one the monolith's PR #810 used). Messages are now exactly as they were; queryability is added on top.

## What's still NOT done (deliberately)

- ~70 long-tail `logger.X(...)` sites where the message has complex expressions like `{len(x)}`, `{x:.2f}`, or function calls. These need per-site judgment to choose sensible field names.
- The 8 sites in benchmarks / tests that use `logging.getLogger`. Out of scope — they don't run in production.

## Test plan

- [x] `uv run python -m compileall -q app/src/context-engine` clean.
- [x] Smoke: `logger.info('demo %s', 42, x=42, src='unit-test')` emits a single line with the positional-formatted message AND the structured kwargs in the tail.
- [x] All three entry points (`http/app`, `cli/main`, `mcp/server`) still import cleanly after migration.
- [ ] Live runtime smoke — left to merger.


Dsantra92:
Closing as not planned for open source.

PR #816, opened 2026-05-27
feat(hatchet): optional durable agent runtime

ASCE-D:
## Summary

Add Hatchet as an opt-in, self-hosted durable runtime for allowlisted agents (debug agent first). Celery remains the default — no behaviour change unless you explicitly enable Hatchet.

- **Runtime layer** — `runtime/` module: `AgentRunSink`, `run_agent_turn`, `RedisStreamSink`, `select_backend`, `hatchet_backend`, `hatchet_worker`. Shared start/chunk/end + cooperative cancel loop extracted from the Celery tasks so both backends run identical application logic.
- **Backend routing** — `conversation_routing._dispatch_agent_run` routes `celery|hatchet` by reading `AGENT_TASK_BACKEND` + `HATCHET_AGENT_ALLOWLIST` (default `debugging_agent`). Resolves the conversation's stored agent before deciding. Fail-closed 503 when Hatchet is selected but unavailable — no silent fallback.
- **Infra (local)** — Docker Compose `hatchet` profile (engine :7077, dashboard :8080); reuses `potpie_postgres` with a separate `hatchet` DB — no RabbitMQ. `scripts/setup_hatchet_local.sh` is idempotent: creates the DB, mints a token, writes `.env.hatchet.local` (gitignored).
- **Makefile** — `hatchet-up`, `hatchet-down`, `hatchet-worker`, `dev-hatchet` (one command: infra + API + Celery + Hatchet worker).
- **Streaming unchanged** — Hatchet runs write to the same Redis stream, so replay and cooperative cancellation work identically.

## How to try it locally

```bash
make dev-hatchet          # one command — boots everything, sets AGENT_TASK_BACKEND=hatchet
# or piecemeal:
make hatchet-up           # boot infra + write .env.hatchet.local
make hatchet-worker       # start the Hatchet agent worker
```

Dashboard: http://localhost:8080 (`admin@example.com` / `Admin123!!`)

See the `# ── Hatchet ──` block in `.env.template` for all config knobs.

## Test Plan
- [ ] `make test-unit` passes (1372 tests, 0 failures)
- [ ] `make dev-hatchet` boots cleanly, dashboard reachable at :8080
- [ ] Send a message on `debugging_agent` — run appears in Hatchet dashboard
- [ ] `make dev` (no Hatchet) still works unchanged

## Notes
[...]

PR #817, opened 2026-05-27
feat(hatchet): route regenerate through backend selection

ASCE-D:
## Summary

**Stacked on:** #816 (feat: optional durable agent runtime) — merge that first.

Regenerate now dispatches through the same Celery/Hatchet backend selector as a normal message, so an allowlisted agent on Hatchet also regenerates on Hatchet instead of always falling back to Celery.

- **`AgentRunInput.operation`** — new discriminator field (`"message"` | `"regenerate"`); one Hatchet event type covers both flows
- **Worker branching** — worker reads `operation` and dispatches to `regenerate_last_message_background` when it's a regenerate
- **`node_ids` parsing hardened** — tolerates `NodeContext` objects (Celery path), raw dicts (Hatchet JSON deserialisation), and plain strings
- **`start_regenerate_task_and_stream`** — new shared helper in `conversation_routing`; the router now calls this instead of going direct to Celery

## Test Plan
- [ ] `make test-unit` passes
- [ ] With `AGENT_TASK_BACKEND=hatchet`: send a message, then hit regenerate — both appear in Hatchet dashboard
- [ ] With default Celery: regenerate still works unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message regeneration workflow for background task execution
  * Enhanced stream replay behavior to display historical events from the beginning when reconnecting
  * Strengthened node ID handling to normalize various input formats consistently across regeneration and new message flows

* **Tests**
  * Added comprehensive unit test coverage for stream replay, regenerate operations, and node ID normalization

[Review Change Stack](https://app.coderabbit.ai/change-stack/potpie-ai/potpie/pull/817?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


yashkrishan (commented):
Review notes from my pass:

- `app/modules/conversations/conversations_router.py`: the new `start_regenerate_task_and_stream(...)` path only covers background streaming regenerates. `stream=false` or `background=false` still goes through the direct controller path and bypasses backend selection. If the PR goal is that regenerate uses the same Celery/Hatchet selector, please route those modes too or explicitly document that they remain direct-only.
- `app/modules/intelligence/agents/hatchet_worker.py`: regenerate now normalizes mixed `node_ids`, but the normal message branch still passes `agent_input.node_ids` directly into `MessageRequest`. `MessageRequest` accepts dict/`NodeContext` values but rejects plain strings, so Hatchet message runs can still fail on string node ids. Please centralize node id normalization and use it for both message and regenerate.
- `app/modules/intelligence/agents/hatchet_worker.py`: please rename the task parameter from `input` to something like `agent_input` or `task_input` to avoid shadowing the Python builtin.
- Please add a worker-level test for the Hatchet regenerate branch, not only the enqueue payload shape, and resolve the failing SonarCloud gate before merge.

yashkrishan:
Two remaining small items from the latest pass:

- `app/modules/intelligence/agents/hatchet_worker.py`: the message operation branch still passes `agent_input.node_ids` directly into `MessageRequest`. Regenerate now handles JSON-shaped node ids, but message runs can still receive plain string node ids from Hatchet payloads and fail validation. Please centralize/normalize node ids and use the same normalized value for both message and regenerate paths.
- `app/modules/intelligence/agents/hatchet_worker.py`: `agent_run_task(input, ctx)` still shadows Python builtin `input`. Please rename the parameter to something like `agent_input` or `task_input` and update the call to `run_agent_via_hatchet(...)`.

yashkrishan (dismissed):

yashkrishan (approved):

PR #830, opened 2026-06-01
Add docker ignore

jdyaberi-pp:
**Summary**
Adds a root .dockerignore so Docker build context stays small and images do not pick up local-only files. Docker does not honor .gitignore, so without this file the full .git tree (often hundreds of MB) was included in context and could be copied into images via COPY . ..

## Summary by CodeRabbit

* **Chores**
  * Improved Docker build configuration to enhance security and reduce build context size.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,83 @@
+# Docker does not use .gitignore — without this file, the full .git directory
+# (~hundreds of MB) is sent as build context and copied into the image via COPY . .
+
+# Version control and CI metadata
+.git
+.github
+
+# Frontend app (separate Docker build under potpie-ui/)
+potpie-ui
+
+# Not required at runtime for API / Celery / Convo images (root only — do not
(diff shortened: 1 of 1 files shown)


yashkrishan (approved):

.dockerignore
```diff
@@ -0,0 +1,76 @@
+# Docker does not use .gitignore — without this file, the full .git directory
+# (~hundreds of MB) is sent as build context and copied into the image via COPY . .
+
+# Version control and CI metadata
+.git
+.github
+
+# Frontend app (separate Docker build under potpie-ui/)
+potpie-ui
```
Dsantra92, line 9: @yashkrishan we are not deploying ui from this image, right? Just double checking.

yashkrishan: correct, potpie-ui is seperate repo, we deploy from there, we should also remove the submodule from here, since we are not using or updating it.

PR #868, opened 2026-06-09
Fix: _process_result always returns truncated result regardless of size

### LINKED ISSUE #724: [Bug] Critical: _process_result always returns None for normal-sized results — silently breaks all code retrieval
## Summary

`_process_result` in `GetCodeFromNodeIdTool` has a missing `return` statement, causing it to return `None` for virtually all results. AI agents silently receive `None` instead of actual code.

## File

`app/modules/intelligence/tools/kg_based_tools/get_code_from_node_id_tool.py`

## Description

The method `_process_result` only has a `return` statement inside an `if len(str(result)) > 80000:` block. For any result under 80,000 characters (the overwhelming majority), the method falls off the end and implicitly returns `None`.

```python
def _process_result(self, result):
    truncated_result = truncate_dict_response(result)
    if len(str(result)) > 80000:
        logger.warning(...)
        return truncated_result
    # ← Missing: return result
    # Method implicitly returns None here!
```

## Impact

- Any AI agent using `get_code_from_node_id` receives `None` instead of the actual code
- Silently breaks **code retrieval for the entire codebase** — QnA, debugging, and test generation pipelines all affected
- No error is raised, making this very hard to debug

## Expected Behavior

The method should return the actual result when it's under the size limit.

## Suggested Fix

Add `return result` as the final line of `_process_result`:

```python
def _process_result(self, result):
    truncated_result = truncate_dict_response(result)
    if len(str(result)) > 80000:
        logger.warning(...)
        return truncated_result
    return result  # ← Add this line
```

manvibuilds:
Fixes #724

## Problem
`_process_result` in `GetCodeFromNodeIdTool` was always returning 
`truncated_result` even for results under 80,000 characters. This meant 
AI agents were receiving unnecessarily truncated responses instead of 
the full result.

## Fix
Return the original `result` when it's under the 80,000 character limit, 
and only return `truncated_result` when truncation is actually needed.

## Changes
- `legacy/app/modules/intelligence/tools/kg_based_tools/get_code_from_node_id_tool.py`
  - Added `return result` for the normal (under size limit) case
  - Moved `return truncated_result` inside the `if` block where it belongs

### DIFF
--- a/legacy/app/modules/intelligence/tools/kg_based_tools/get_code_from_node_id_tool.py
+++ b/legacy/app/modules/intelligence/tools/kg_based_tools/get_code_from_node_id_tool.py
@@ -219,7 +219,8 @@ class GetCodeFromNodeIdTool:
                 node_id=node_id,
                 project_id=project.id,
             )
-        return truncated_result
+            return truncated_result
+        return result
 
     @staticmethod
     def _get_relative_file_path(file_path: str) -> str:


manvibuilds:
The failing tests aren't related to this fix they're all throwing FileNotFoundError for a missing linear/issue_detail.json test data file, which seems like a pre-existing issue in the repo. My change only touches the legacy tool file. which passes all the test.

PR #871, opened 2026-06-09
Fix context-engine CI regression for missing fixtures and flaky OAuth

Nih1tGupta:
## Summary
- Skip benchmark smoke test when the `universe/acme` fixture corpus is not shipped
- Add `tests/data/linear/issue_detail.json` for Linear issue resolver unit tests
- Stabilize OAuth callback tests (start server before HTTP hit) to avoid CI races
- Align setup conformance test with default pot name `foo-pot`
## Test plan
- [x] Previously failing regression tests pass locally
- [ ] `potpie-context-engine` CI green on Python 3.11/3.12/3.13


Dsantra92 (approved):
lgtm

nndn (approved):

PR #882, opened 2026-06-13
Fix interactive setup source deferral

### LINKED ISSUE #881: Default pot flow doesn't take in the new pot if one already exists
There is an edge case in interactive setup where repo registration can happen before the first pot is created.

In the setup flow, first-pot creation is deferred to the post-setup wizard so the user can choose the pot name. However, the setup orchestrator can still run the `source` step before that wizard step. If the user already has an active pot, the repo may be registered against that existing pot instead of the new first pot created during setup.

## Reproduction Steps

1. Start with a Potpie state that already has an active pot.
2. Run interactive setup with a repo:

   ```bash
   potpie setup --repo .
   ```

3. Use the Rich interactive path. Do not use `--json` or `--yes`.
4. Continue until the wizard asks for the first pot name.
5. Check the sources on the previously active pot.

## Expected Behavior

The old active pot should not receive the repo source.

The repo should only be registered after the wizard creates and activates the new first pot.

## Actual Behavior

The repo can be registered against the old active pot because it is active when the orchestrator runs the `source` step.

Dsantra92:
## Summary

- defer repo/source registration when interactive setup defers first-pot creation
- keep the setup preview reason aligned with the runtime behavior
- add regression coverage for an existing active pot receiving no repo source during deferred setup

## Why

Interactive setup creates the first pot later in the post-setup wizard. If source registration runs before that wizard step, a user with an existing active pot can have the repo attached to the old pot instead of the newly named first pot.

Fixes #881.

## Verification

- uv run --package potpie-context-engine pytest potpie/context-engine/tests/unit/test_setup_defer_pot.py -q
- uv run --package potpie-context-engine pytest potpie/context-engine/tests/unit/test_setup_*.py -q
- uv run --package potpie-context-engine pytest potpie/context-engine/tests/conformance/test_host_shell_end_to_end.py::test_setup_orchestrator_provisions_and_creates_default_pot potpie/context-engine/tests/conformance/test_host_shell_end_to_end.py::test_setup_is_idempotent potpie/context-engine/tests/conformance/test_host_shell_end_to_end.py::test_setup_dry_run_executes_nothing potpie/context-engine/tests/conformance/test_host_shell_end_to_end.py::test_setup_preview_classifies_owners_and_host_gated_hardness -q
- uv run --package potpie-context-engine ruff check potpie/context-engine/application/services/setup_orchestrator.py potpie/context-engine/tests/unit/test_setup_defer_pot.py
- uv run --package potpie-context-engine basedpyright --level error potpie/context-engine/application/services/setup_orchestrator.py potpie/context-engine/tests/unit/test_setup_defer_pot.py
- git diff --check
- isolated CLI smoke: potpie --json setup --repo . --agent claude --yes

### DIFF
--- a/potpie/context-engine/application/services/setup_orchestrator.py
+++ b/potpie/context-engine/application/services/setup_orchestrator.py
@@ -90,6 +90,8 @@ def _skip_reason(step: str, plan: SetupPlan) -> str | None:
         return "in-process host — no detached daemon / service unit to manage"
     if step == "source" and not plan.repo:
         return "no --repo provided"
+    if step == "source" and plan.defer_default_pot:
+        return "deferred until post-setup first pot"
     if step == "pot.default" and plan.defer_default_pot:
         return "named in post-setup wizard"
     if step == "skills" and plan.defer_skills:
@@ -194,11 +196,7 @@ class DefaultSetupOrchestrator:
             self._step("auth", hard("auth"), self.auth.init_local),
             self._step("source", hard("source"), lambda: self._source(plan)),
             *(
-                [
-                    self._step(
-                        "skills", hard("skills"), lambda: self._skills(plan)
-                    )
-                ]
+                [self._step("skills", hard("skills"), lambda: self._skills(plan))]
                 if not plan.defer_skills
                 else []
             ),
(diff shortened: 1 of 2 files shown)


shmbhvi101 (approved):
LGTM. Initially the pot creation step was during pre-setup and `source` correctly attached the right repo to the newly created default pot. This behaviour was changed as we pushed the pot creation post-setup for UX reasons but `source` behaviour was not altered accordingly. This PR targets that bug correctly and we're good to merge this.

PR #891, opened 2026-06-16
smooth GitHub login cancel flow, auto-open countdown, and setup skip when authenticated

shmbhvi101:

PR #880, opened 2026-06-12
fix(cli): smooth GitHub login cancel flow, auto-open countdown, and setup skip when authenticated

BrhKmr23:
## Summary
Fixes https://github.com/potpie-ai/potpie/issues/879
- Improve GitHub device-flow login UX: after showing the verification code, show an inline countdown (`Press Enter to open now, or GitHub opens in Ns`) and auto-open the browser after 10 seconds instead of blocking on a separate confirm prompt.
- Handle user cancellation cleanly on Ctrl+C / Click Abort during login — print `GitHub login cancelled.`, exit with code 130, and avoid tracebacks, token polling, or partial credential writes.
- Skip the post-setup “Would you like to log in to GitHub now?” prompt when GitHub is already authenticated (shows `GitHub already connected as <login>; skipping login.`).
- Roll back the keychain token if integration metadata write fails after a successful token store, preventing orphaned secrets.
- Use `msvcrt` on Windows so Enter-to-open works (Unix uses `select` on stdin).

## Test plan
Screenshot 2026-06-12 at 5 23 29 PM

- [x] `uv run pytest tests/unit/test_github_cli_auth.py tests/unit/test_credentials_store.py tests/unit/test_setup_post_auth_prompts.py`
- [ ] `potpie github login` — verify countdown, Enter opens browser immediately, auto-open after 10s
- [ ] Ctrl+C at the countdown prompt — verify clean cancel message and exit code 130
- [ ] `potpie setup` with GitHub already logged in — verify login prompt is skipped
- [ ] `potpie setup` without GitHub credentials — verify login prompt still appears
- [ ] (Windows) verify Enter-to-open works during the countdown prompt

PR #907, opened 2026-06-22
fix(cli): post-setup integration login cancel flow + onboarding analytics (cg-01)

nndn:
Root PR **cg-01** of the `cg-NN` rollout. Originally tracked as "merge `feat/CG-demo` (#891) + cg-01", but investigation showed **#891 is already fully merged into `main`** (5/7 files byte-identical, `github_commands.py` a superset, `setup_ux.py` already has the cancel-flow/auto-open/setup-skip features — landed via #880/#887/#888). So this PR carries only **cg-01's unique, not-yet-merged work**, rebased onto `main` and self-contained.

## Changes
- **Post-setup integration picker** hardened against cancellation: each provider login skips cleanly on Ctrl+C / EOF / `click.Abort` instead of aborting the whole setup, prints `Skipped <X> — connect later with \`potpie <x> login\``, and records a `user_aborted` skip event.
- **Jira/Confluence → single Atlassian** option in the picker.
- Surface config/data paths in the completion summary; route integration auth through shared onboarding analytics.

## Bug fixed (would have failed CI)
- **`NameError: click`** in `commands/_common.py`: the `contract()` error boundary special-cased `click.Abort` but never imported `click`, so any non-Abort exception reaching that branch raised `NameError` instead of being captured. Added `import click` (+2 regression-covered sentry tests).

## Scope notes
- Trimmed cg-01's forward dependency on `adapters/inbound/cli/repo_location.py` + `pots.set_repo_default` (those belong to **cg-07**, not on `main` yet). `_register_repo_source` reverts to plain `add_source(location=repo)`.
- Does **not** touch the 3 cg-04-diverged files (`graph_workbench*`, `skill_manager`) — no revert risk.

## Test
- Full CE unit suite **green in a clean CI-equivalent worktree: 1640 passed, 0 failed**.
- Note: locally, `test_product_analytics_dispatcher_flushes_queued_events` fails only when `POTPIE_POSTHOG_API_KEY` is set (repo-root `.env`) — pre-existing on `main` too, env-only, not a regression.

Supersedes #891 (already in main).


potpie/context-engine/tests/unit/test_usage_analytics.py
```diff
@@ -104,22 +104,6 @@ def test_context_activation_also_records_usage(fake_sink: _FakeSink) -> None:
     assert fake_sink.events[-1].properties["item_count"] == 2
 
 
-def test_ingest_scan_activation_also_records_usage(fake_sink: _FakeSink) -> None:
```
shmbhvi101, line 107: why is this deleted?

shmbhvi101 (approved):
LGTM

PR #929, opened 2026-06-24
feat(cli): source status no-ID summary and doctor effective repo pot  (POT-1763 and POT-1779)

Nih1tGupta:
## Summary

- **Audit 10:** `potpie source status` now works without a source ID and returns a per-pot summary (`kind`, `location`, `repo_default`, `claim_count`, `registration_only`, `ingestion_status`, `recommended_next_action`). Passing a source ID returns the same enriched single-source shape instead of the old `{id, status, name}` stub.
- **Audit 26:** `potpie doctor` now reports `active_pot`, `repo_default_pot`, and `effective_current_repo_pot` so diagnostics align with how `source list` and `graph status` resolve the pot in the current repo. Plain-text output includes a `repo: <identity> → <pot>` line when inside a git repo.

## Test plan

- [x] `pytest tests/unit/ -q` — 1801 passed
- [x] `potpie --json source status` — per-pot summary with enriched `sources[]`
- [x] `potpie --json source status <source-id> --pot <pot>` — enriched single-source row
- [x] `potpie --json doctor` — includes `active_pot`, `repo_default_pot`, `effective_current_repo_pot`
- [x] `potpie doctor` — human output shows repo routing line
- [x] After `potpie pot default set --repo current <pot>`, `repo_default_pot` and `effective_current_repo_pot` update correctly
- [x] New unit tests in `test_source_cli_contract.py` and `test_cli_bootstrap_status.py`


potpie/context-engine/adapters/inbound/cli/commands/bootstrap.py
```diff
@@ -270,6 +272,13 @@ def doctor() -> None:
             pot_id = getattr(pot, "pot_id", "") if pot is not None else ""
             readiness = host.backend.mutation.readiness(pot_id)
             daemon_status = host.daemon.status()
+
+            repo_identity = current_repo_identity_for_cli()
+            default_pot_id = repo_default_pot_id(host, repo_identity)
+            effective_current_repo_pot = (
```
yashkrishan: This does not fully match the current repo pot resolution used elsewhere. It only considers the repo default and then falls back to the active pot, but `resolve_pot_scope()` / `source list` also resolve a single linked repo pot before falling back to active. In a repo linked to pot `p1`, with active pot `p2` and no repo default, `source list`/graph commands resolve to `p1` while `doctor` would report `effective_current_repo_pot=p2`. Consider deriving this from `resolve_pot_scope(host)` (or equivalent shared helper) so doctor reports the same effective pot as the CLI actually uses.

nndn (approved):

PR #949, opened 2026-06-26
feat(cli): Add Gitbucket integration (POT-1722)

Nih1tGupta:
## Summary

Adds GitBucket CLI integration for self-hosted instances using a **personal access token** (no OAuth — GitBucket has no central auth server).

## Commands

| Command | Description |
|---------|-------------|
| `potpie gitbucket login` | Connect via PAT (interactive) |
| `potpie gitbucket login --host <url> --token <pat>` | Non-interactive login |
| `potpie gitbucket login --force` | Re-authenticate |
| `potpie gitbucket logout` | Remove stored credentials |
| `potpie gitbucket repos` | List accessible repos |
| `potpie gitbucket repos -n 10` | Limit repo count |
| `potpie status` / `potpie status --verify` | Shows GitBucket connection + verifies token |

Compat (deprecated): `potpie auth gitbucket login|logout|repos`

## How it works

1. **Login** — user provides host URL (e.g. `http://localhost:8080`) and PAT from Account Settings → Applications.
2. **Verify** — CLI calls `GET {host}/api/v3/user` with `Authorization: token {PAT}`.
3. **Store** — token in `integration_secrets.json`; host + account metadata in `credentials.json`.
4. **Repos** — `GET {host}/api/v3/user/repos` lists repos the user can access.
5. **Logout** — clears GitBucket credentials from local storage.

Supports subpath deployments (e.g. `https://git.company.com/gitbucket`).

## Test plan

- [ ] `potpie gitbucket login` + `repos` against local GitBucket
- [ ] `potpie status --verify`
- [ ] `potpie gitbucket logout`
- [ ] `pytest tests/unit/test_gitbucket_cli.py -q`
Screenshot 2026-06-25 at 3 29 44 PM


potpie/context-engine/adapters/outbound/cli_auth/provider_config.py
```diff
@@ -31,6 +31,15 @@
     f"{ATLASSIAN_API_GATEWAY}/oauth/token/accessible-resources"
 )
 
+# GitBucket PATs are created at /{username}/_application (Account Settings → Applications).
+GITBUCKET_TOKEN_PAGE_SUFFIX = "_application"
```
jdyaberi-pp, line 43: Fixed: GitBucket PATs live at `/{username}/_application` (Account Settings → Applications), not GitHub's `/settings/tokens`. `gitbucket_token_page_url` now builds the correct path and interactive login prompts for the username first.

potpie/context-engine/adapters/outbound/cli_auth/gitbucket_read_client.py
```diff
@@ -0,0 +1,138 @@
...
+    token_provided = token is not None
+    if host_provided != token_provided:
+        raise GitBucketReadError(
+            "GitBucket host_url and token must be provided together."
+        )
+    if not host_provided and not token_provided:
+        host_url, token = _load_credentials()
+
+    try:
+        ensure_gitbucket_pat_transport_allowed(host_url)
```
jdyaberi-pp, line 104: Fixed: apply `ensure_gitbucket_pat_transport_allowed` before sending the stored PAT so `potpie gitbucket repos` cannot transmit tokens over plain HTTP to non-loopback hosts when verify would reject the same host.

potpie/context-engine/adapters/outbound/cli_auth/gitbucket_read_client.py
```diff
@@ -0,0 +1,138 @@
...
+    try:
+        url = f"{api_base}/user/repos?per_page={limit}"
+        data = _get_json(url, headers=headers, http=http)
+    finally:
+        if owns:
+            http.close()
+
+    if not isinstance(data, list):
+        raise GitBucketReadError(
+            "GitBucket API returned an unexpected response format for /user/repos."
```
jdyaberi-pp, line 121: Fixed: raise `GitBucketReadError` when `/user/repos` returns HTTP 200 with a non-array body instead of returning an empty list that masks API/compatibility issues.

jdyaberi-pp (commented):
Addressed three Bugbot findings in 86b1087d:

1. **Token page URL** — switched from GitHub-style `/settings/tokens` to GitBucket `/{username}/_application`; interactive login now prompts for username before opening the browser.
2. **HTTP PAT guard** — `list_gitbucket_repos` now calls `ensure_gitbucket_pat_transport_allowed` so repo listing matches verify/status transport policy.
3. **Non-list API responses** — `/user/repos` payloads that are not JSON arrays now raise `GitBucketReadError` instead of silently returning an empty list.

potpie/cli/auth/auth_commands.py
```diff
@@ -423,7 +426,7 @@ def _run_linear_oauth_flow(*, force: bool = False, add: bool = False) -> None:
 
 def _integration_auth_provider(provider: str) -> IntegrationAuthProvider:
     key = provider.strip().lower()
-    if key in {"linear", "atlassian", "jira", "confluence"}:
+    if key in {"linear", "atlassian", "jira", "confluence", "gitbucket"}:
```
yashkrishan, line 429: check if gitlab needs to be added here, since it has been merged already

yashkrishan (commented):
Reviewed current diff; adding inline comments for the remaining issues.

potpie/cli/auth/gitbucket_commands.py
```diff
@@ -0,0 +1,444 @@
...
+
+
+def _open_computed_token_page(token_page: str, *, as_json: bool) -> bool:
+    if not as_json:
+        print_plain_line(f"Opening {token_page} ...", as_json=False)
+    opened = webbrowser.open(token_page, new=1)
+    if as_json:
+        print_json_blob(
+            {
+                "ok": True,
```
yashkrishan: This prints an `ok: true` JSON object before the login is actually complete. A script reading JSON output may treat this as success even if token verification or saving fails later. Can we either avoid `ok: true` for this intermediate token-page event, or include the token page URL in the final/error payload instead?

potpie/cli/auth/gitbucket_commands.py
```diff
@@ -0,0 +1,444 @@
...
+        print_json_blob(
+            {"ok": True, "provider": "gitbucket", "count": len(repos), "repos": repos},
+            as_json=True,
+        )
+        return
+
+    for repo in repos:
+        visibility = "private" if repo.get("private") else "public"
+        print_plain_line(
+            f"{repo.get('full_name')}\t{visibility}\t{repo.get('default_branch') or ''}",
```
yashkrishan: These repo fields come from GitBucket, but `print_plain_line` parses Rich markup by default. Please escape `full_name` and `default_branch`, or print this line with `markup=False`, so repo names with markup-like text cannot change the terminal output.

potpie/cli/auth/auth_commands.py
```diff
@@ -89,8 +89,11 @@
     "jira",
     "confluence",
     "gitlab",
+    "gitbucket",
```
yashkrishan, line 92: Since this adds `gitbucket` to the provider list alongside the recently merged `gitlab` provider, please also update the user-facing `auth logout` help text below. It still says only `github, linear, jira, or confluence`, so users will not see that `gitlab` and `gitbucket` are accepted there too.

potpie/cli/auth/auth_commands.py
Nih1tGupta: GitLab intentionally does not belong in _integration_auth_provider. That list is only for providers that log in through the shared setup/onboarding helper (run_integration_login). GitLab follows the GitHub-style pattern and uses its own top-level command (potpie gitlab login), so it is wired via _ALL_PROVIDERS, gitlab_app, and gitlab_logout_impl() instead. Adding "gitlab" here would require also extending IntegrationAuthProvider and run_integration_login() with a GitLab branch, which is out of scope for this PR

yashkrishan (approved):

PR #950, opened 2026-06-26
Add pre-merge CLI journey integration gate

jdyaberi-pp:
## Summary
- add a new context-engine integration test that simulates a fresh user journey (clone, install, setup, source registration, graph mutation/read checks)
- keep the test isolated with temp HOME/XDG directories and non-interactive CLI setup
- add a dedicated `premerge-cli-journey` GitHub Actions job to run this journey test on PRs

## Test plan
- [x] `uv run --project potpie/context-engine ruff check potpie/context-engine/tests/integration/test_premerge_cli_journey.py`
- [x] `uv run --project potpie/context-engine pytest potpie/context-engine/tests/integration/test_premerge_cli_journey.py -v`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -113,7 +113,9 @@ jobs:
           version: "0.10.7"
 
       - name: Set up Rust toolchain
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8
+        with:
+          toolchain: stable
 
       - name: Install dependencies
         run: uv sync --frozen --all-packages --no-cache
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -112,6 +112,9 @@ jobs:
         with:
           version: "0.10.7"
 
+      - name: Set up Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
       - name: Install dependencies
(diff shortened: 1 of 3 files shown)


yashkrishan:
Two things to address before we rely on this as a merge gate:

**1. Critical — the dedicated job isn’t actually running the test**

The `premerge-cli-journey` CI job is green, but the test is **skipped** (not passed). In the latest run it finished in ~0.5s with `SKIPPED` — so we’re not exercising clone → install → setup → graph mutate/read today.

Root cause: the test isolates `HOME` to a temp directory, but `RUSTUP_HOME` isn’t carried over when GitHub Actions only sets `CARGO_HOME`. Rustup then can’t find a toolchain in the isolated env, the Rust preflight fails, and the test skips. The same skip happens in the regular `potpie-context-engine` matrix jobs too.

**Fix:** preserve `RUSTUP_HOME` before overriding `HOME` (e.g. fall back to the real home’s `.rustup`), and ideally fail instead of skip in the dedicated job so a broken gate is obvious.

---

**2. Question — should this run on every push to `main`?**

Right now the workflow runs on PRs **and** every push to `main`. Each run does a full clone + `uv sync` inside the test (slow/heavy; job timeout is 45 min). That’s useful for catching regressions after merge, but it could add a lot of CI cost over time.

**Do we want this job PR-only, or is running it on every `main` push intentional?**

yashkrishan (approved):

PR #970, opened 2026-06-29
Add bounded UI dev port fallback

### LINKED ISSUE #969: Make local Potpie UI dev server prefer 3000 with bounded fallback ports
The graph explorer frontend currently uses Vite’s default dev server behavior, so `npm run dev` starts on `5173` and Vite may choose the next available port if that port is occupied. For Potpie local development, we want the UI dev server to use the expected 3000-series URL instead.

Implement a clear, bounded port selection policy for the local UI dev server:

- Prefer `3000`.
- If `3000` is unavailable, try `3001`, then increment sequentially.
- Stop at a clear maximum, e.g. `3099`.
- If no port in the range is available, fail with an actionable error that tells the developer which range was tried.
- Keep the selected URL visible in the dev server output.
- Do not change the packaged `potpie ui` daemon behavior unless needed; the daemon-served UI should continue to use daemon discovery.

Implementation notes:

- Avoid a separate “check then bind” race where possible.
- If this is implemented through Vite config or a small dev launcher, keep the behavior obvious and local to the graph explorer frontend.
- Preserve `/ui/` as the frontend base path.
- Preserve the `/ui/api` proxy behavior to the daemon.

Acceptance criteria:

- Running the local frontend dev server uses `http://localhost:3000/ui/` when port 3000 is free.
- If 3000 is occupied, it uses the next free port, such as `3001`, `3002`, etc.
- The fallback stops at the configured max port and reports a clear failure if the range is exhausted.
- The daemon-served `potpie ui` flow still opens the daemon discovery URL under `/ui`.
- Existing frontend build still works with `make ui-build` / `npm run build`.
- Add or update focused tests where practical, or document why this behavior is covered by the chosen Vite/dev launcher mechanism.

Dsantra92:
## Summary

- Replaces the graph explorer frontend dev script with a small Vite launcher that prefers port 3000 and retries sequentially through 3099.
- Keeps the bounded retry loop atomic by asking Vite to bind each candidate with `strictPort: true` instead of doing a separate preflight port scan.
- Adds focused Node tests for range validation, Vite bind-error detection, fallback behavior, exhaustion errors, unsupported extra dev args, and non-port failure handling.
- Wires the port tests into `npm run build`, so `make ui-build` now runs them before producing the bundle.

Fixes #969.

## Review Fixes

- Extra `npm run dev -- ...` Vite flags now fail loudly instead of being silently ignored by the custom launcher.
- The Vite v5 strict-port error text match is documented in code as a compatibility boundary.
- The new port-policy tests are now part of the normal frontend build path.

## Validation

- `npm test`
- `npm run build`
- `node scripts/dev.mjs --host 0.0.0.0` fails loudly for unsupported args
- runtime smoke: `scripts/dev.mjs` selects port 3000 when free
- runtime smoke: with 3000 occupied, `scripts/dev.mjs` selects port 3001
- `git diff --check`
- `make ui-build`
- thermo-nuclear code-quality re-audit on final diff: no file-size, boundary, spaghetti-branching, silent-arg-drop, or test-discoverability blockers remain

## Notes

`make ui-build` runs `npm install` and reports existing npm advisories in the dependency tree. This PR leaves dependency upgrades out of scope for the port-selection change.

### DIFF
--- a/potpie/context-engine/adapters/inbound/http/ui/frontend/package.json
+++ b/potpie/context-engine/adapters/inbound/http/ui/frontend/package.json
@@ -4,7 +4,8 @@
   "version": "0.1.0",
   "type": "module",
   "scripts": {
-    "dev": "vite",
+    "dev": "node scripts/dev.mjs",
+    "test:dev-port": "node --test scripts/dev-port.test.mjs",
     "build": "tsc -b && vite build",
     "preview": "vite preview"
   },
--- a/potpie/context-engine/adapters/inbound/http/ui/frontend/package.json
+++ b/potpie/context-engine/adapters/inbound/http/ui/frontend/package.json
@@ -5,8 +5,9 @@
   "type": "module",
   "scripts": {
     "dev": "node scripts/dev.mjs",
+    "test": "npm run test:dev-port",
     "test:dev-port": "node --test scripts/dev-port.test.mjs",
-    "build": "tsc -b && vite build",
+    "build": "npm run test:dev-port && tsc -b && vite build",
     "preview": "vite preview"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


ASCE-D (approved):
LGTM

PR #1006, opened 2026-07-13
Fix(cli): detect resolve intent and report intent_source

### LINKED ISSUE #996: CLI: resolve defaults to feature intent without detection
### Area

Other

### Potpie version

potpie v2.0.0b3

### Install source

```bash
uv tool install potpie
```

### System

macOS

### Coding agent and model

_No response_

### Command or workflow

```bash
# Naive resolve (no --intent)
potpie resolve "what changed recently in the auth service?"
potpie --json resolve "the payment webhook is throwing a 500 error and crashing"

# Compare with explicit intent
potpie --json resolve "what changed recently in the auth service?" --intent debugging
potpie --json resolve "some task" --intent feature
```

## Problem

When `potpie resolve "<task>"` is run **without** the `--intent` flag, the CLI silently defaults to `intent=feature` and displays it as though the intent was detected. However, **no intent detection is actually performed**.

## Observed Behavior

- Different task descriptions can return the same claim set.
- The output header displays:

  ```text
  intent=feature
  confidence=low
  ```

  even though no intent classification was executed.

- Queries about recent changes can return **zero timeline items**, because the default `feature` intent does not include timeline-related retrieval.
- Providing `--intent` explicitly produces significantly better results than relying on the default behavior.
- `confidence=low` may appear even when the retrieved results are highly relevant because the confidence value is **coverage-based**, not a measure of match quality.

## Expected Behavior

- Do **not** silently default to `intent=feature` when `--intent` is omitted.
- Detect the intent from the task description, or transparently fall back to a neutral value such as `unknown` if the intent cannot be determined.
- Clearly indicate how the intent was selected, for example:
  - `explicit`
  - `detected`
  - `default`
- Route debugging and recent-change queries to retrieval strategies that include relevant families (such as `timeline`) instead of always using the `feature` retrieval shape.

## Impact

The current behavior causes naïve `potpie resolve` calls to use an inappropriate retrieval strategy while giving the impression that the intent was intentionally selected. As a result:

- AI agents may retrieve incomplete or irrelevant context.
- Additional tokens are wasted on follow-up queries.
- Users and agents can lose trust in the `confidence` metadata, since `confidence=low` reflects retrieval coverage rather than the quality or correctness of the match.

### Source or build details

_No response_

AdityaPainuli:
Closes potpie-ai/potpie#996

## Problem

`potpie resolve "<task>"` without `--intent` silently hardcodes `intent=feature` and prints it as if it was detected. No detection ever runs, so debugging and recent-change tasks get the feature retrieval shape (no `timeline` family), and the output misleads users into thinking the intent was classified.

## What changed

* Added `detect_context_intent(task)` in `domain/agent_context_port.py`: a deterministic, dependency-free keyword/phrase matcher (case-insensitive, word-boundary) seeded from the trigger language in each recipe's `"when"` field. Multi-match ties break via a documented priority order. Returns `None` when nothing matches, favoring no guess over a wrong guess, mirroring the `ontology_classifier` discipline.
* CLI `resolve`: `--intent` default changed from `"feature"` to `None`. Effective intent now resolves explicit → detected → default (`unknown`), following the `(value, resolved_via)` pattern from `resolve_pot_id`/`resolve_pot_scope`.
* The selection provenance rides the existing `ResolveRequest.metadata` → `AgentEnvelope.metadata` passthrough as `intent_source` (`explicit` | `detected` | `default`). No downstream signature changes.
* Both outputs surface it: JSON gets an `intent_source` key, the human header shows `intent=<x> (source=<...>)`.

Vocabulary, recipes, `normalize_context_intent`, `DEFAULT_INTENT_INCLUDES`, and the MCP/HTTP surfaces are untouched. Correct detection routes recent-change/debugging tasks to timeline-bearing includes automatically via the existing intent→includes mapping.

## Behavior

| Command | Before | After |
| -- | -- | -- |
| `resolve "webhook throwing a 500 error"` | `intent=feature` | `intent=debugging (source=detected)` |
| `resolve "what changed recently in auth?"` | `intent=feature`, zero timeline items | `intent=review (source=detected)`, timeline included |
| `resolve "xyz"` (no match) | `intent=feature` | `intent=unknown (source=default)` |
| `resolve "..." --intent docs` | `intent=docs` | `intent=docs (source=explicit)` |

## Testing

* 18 new unit tests: representative task→intent mappings, recent-change phrasing selects a timeline-bearing intent, empty/ambiguous input returns `None`, word-boundary anti-false-match ("latest" must not trigger `test`), CLI precedence, and `intent_source` in both output formats.
* Full unit suite: 2063 passed, 1 skipped, 0 failures.
* `ruff check` and `ruff format` clean (repo pre-commit config, v0.14.3).


potpie/context-core/src/potpie_context_core/agent_context_port.py
```diff
@@ -172,6 +173,131 @@
...
+    "feature",
+)
+
+_INTENT_MATCHERS: dict[str, re.Pattern[str]] = {
+    intent: re.compile(
+        r"|".join(r"\b" + re.escape(kw) + r"\b" for kw in keywords),
+        re.IGNORECASE,
+    )
+    for intent, keywords in _INTENT_KEYWORDS.items()
+}
```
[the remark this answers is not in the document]
AdityaPainuli: Valid, resolved now. Removed the drift possibility instead of guarding it. So the two structures can't diverge.

potpie/context-core/src/potpie_context_core/agent_context_port.py
```diff
@@ -172,6 +173,119 @@
...
+
+    Deterministic, case-insensitive, word-boundary keyword match. When the
+    task triggers more than one intent, :data:`_INTENT_DETECTION_PRIORITY`
+    breaks the tie (most specific / highest-risk first). Returns ``None`` when
+    nothing matches — favouring "no guess" over a wrong guess — which the CLI
+    surfaces as the neutral ``unknown`` intent with ``intent_source=default``.
+    """
+    if not task or not task.strip():
+        return None
+    for intent in _INTENT_DETECTION_PRIORITY:
```
FarbodDaneshjoo, line 290: Nice work! I noticed there's good coverage for representative mappings and the fallback behavior, but I couldn't find a regression test for a task matching multiple intents.

Since `detect_context_intent()` relies on `_INTENT_DETECTION_PRIORITY` to resolve ties, would it make sense to add a regression test for a multi-match phrase? That would help ensure the documented priority order remains stable over time.

AdityaPainuli: Good catch, added. It pins four multi-match phrases ( security > debugging > operations > review > feature), and each case also guards-asserts the lower-priority intents genuinely match the phrase, so a future keyword change can't silently reduce the test to a single-match.

FarbodDaneshjoo: Thanks for adding it! I like the extra guard assertions—they make the priority behavior much clearer and help ensure the test stays meaningful over time.

AdityaPainuli:
@BrhKmr23 solved the conflicts, feel free to review the PR now.

potpie/cli/commands/query.py
```diff
@@ -128,6 +153,7 @@ def _envelope_payload(env) -> dict[str, object]:
     return {
         "pot_id": env.pot_id,
         "intent": env.intent,
+        "intent_source": dict(env.metadata).get("intent_source", "default"),
```
jdyaberi-pp: These helpers format both `resolve` and `search`. Search envelopes contain no `intent_source`, so this fallback makes every `potpie search` JSON response report `intent_source="default"`, while `_envelope_human` below similarly shows `(source=default)`, even though search performs no intent selection. This is misleading and changes an unrelated command’s output. Could we only render the field when the metadata actually contains it, or use resolve-specific formatting?

jdyaberi-pp (commented):
I found one output-path issue to address.

potpie/cli/commands/query.py
AdityaPainuli: Good catch, fixed. Went with your first suggestion: the formatters now render intent_source only when the envelope metadata actually carries it. `resolve` output is unchanged (it always sets the field), and `search` output is back to its pre-PR shape. Also, added two tests asserting the field never leaks into search-shape envelopes.

jdyaberi-pp (approved):

AdityaPainuli:
@jdyaberi-pp can you please approve and merge these changes. Thanks!

PR #1032, opened 2026-07-28
Prevent stale entity labels during re-type upserts

rishita2023:
Entity-key prefixes now determine the canonical type, conflicting labels are removed before persistence, and label inference cannot reintroduce them. Adds quality detection and an explicit repair command for existing polluted FalkorDB/Neo4j nodes, with unit and live FalkorDB Lite coverage.


rishita2023:
@coderabbitai review

potpie/context-core/src/potpie_context_core/reconciliation_validation.py
```diff
@@ -57,9 +57,16 @@ def validate_reconciliation_plan(
         config or current_reconciliation_config() or reconciliation_config_from_env()
     )
     definition = definition or DEFAULT_GRAPH_DEFINITION
+    plan.ontology_downgrades.clear()
+    canonicalize_reconciliation_plan(
+        plan,
+        enforce_label_coherence=config.infer_canonical_labels,
+        entity_types=definition.entity_types,
```
Dsantra92, line 64: Keep key-prefix label coherence independent of label inference. `infer_canonical_labels` controls classifier enrichment; using it for `enforce_label_coherence` means `CONTEXT_ENGINE_INFER_LABELS=0` lets `environment:*` entities retain conflicting canonical labels (for example, both `Service` and `Environment`), so this PR's core invariant is not enforced. Pass `enforce_label_coherence=True` here, retain the flag only around `enrich_reconciliation_plan_entity_labels`, and add a regression test with inference disabled.

potpie/context-core/src/potpie_context_core/workbench_service.py
```diff
@@ -2640,6 +2653,95 @@ def _quality_projection_drift(
...
+                pot_id=pot_id,
+                subgraph_in=(subgraph,) if subgraph else (),
+                limit=_quality_scan_limit(limit),
+            ),
+        )
+        for node in graph_slice.nodes:
+            entity_keys.add(node.key)
+            labels_by_key[node.key] = tuple(node.labels)
+    except CapabilityNotImplemented:
+        pass
```
Dsantra92: Do not report a complete/`ok` label-drift scan when inspection coverage is incomplete. This handler silently discards `CapabilityNotImplemented`, and the success path also ignores `graph_slice.truncated`. Neo4j inspection is unimplemented, so the fallback sees only claim-referenced keys; FalkorDB caps slices at 2,000 nodes. Drift in isolated or later nodes can therefore be missed while `_quality_status` returns `ok`. Use paginated all-entity enumeration, or propagate unsupported/truncated coverage so the result is `partial`, with regression tests for both cases.

Dsantra92 (changes requested):
Requesting changes for two P1 correctness gaps: key-prefix label coherence is disabled when label inference is off, and the entity-label drift report can return a clean result after an incomplete scan. Details inline.

potpie/context-core/src/potpie_context_core/reconciliation_validation.py
rishita2023: Addressed, I now enforce key-prefix label coherence unconditionally during reconciliation validation (`enforce_label_coherence=True`) and keep `infer_canonical_labels` only around classifier-based enrichment. I also added a regression test for the inference-disabled path to verify `environment:*` entities no longer retain conflicting canonical labels.

potpie/context-core/src/potpie_context_core/workbench_service.py
rishita2023: Addressed, We now surface incomplete inspection coverage in the label-drift report instead of returning a misleading clean result, and I added regression tests for both unsupported and truncated inspection.

rishita2023 (commented):
Fixed both. Key-prefix label coherence is now enforced independently of classifier inference, with a regression test for the inference-disabled path, and the label-drift quality report now propagates incomplete inspection coverage rather than reporting a clean `ok` when inspection is unsupported or truncated, with regression tests covering both cases.

Dsantra92 (approved):

PR #1070, opened 2026-08-31
fix: close the 2.0.1 release-test findings

nndn:
Someone ran a full test sweep against potpie 2.0.1 after it went up on PyPI and
found 20 problems. This fixes all of them.

I reproduced every one first — against a throwaway daemon and throwaway pots, so
nothing touched a real setup — then fixed it, then checked it was actually gone.

---

### Upgrading potpie while the daemon is running bricks the CLI

If you had a daemon running and then upgraded, the new potpie couldn't talk to
the old daemon. Fine so far — except it also couldn't stop it, restart it, or
replace it. Every suggestion it gave you led nowhere:

| you run | you get |
|---|---|
| `potpie status` | "discovery is unavailable — run `potpie daemon restart`" |
| `potpie daemon restart` | "Unexpected internal error." |
| `potpie daemon stop` | "refusing to signal the recorded process" |
| `potpie daemon start` | "daemon already running" |
| `potpie daemon status` | "up=True" — which contradicts `status` |

The only way out was to find the process ID yourself and `kill` it. Nothing in
the CLI told you that. This is what *every* existing user hit when they upgraded.

**Now:** potpie recognises the old daemon and says so in plain words — "the
runtime record was written by an older potpie". `daemon restart` genuinely
recovers: it shuts the old daemon down and starts a fresh one. `daemon status`
admits the daemon is there but unusable, and prints the `kill` command as a last
resort. `daemon stop` offers `--force`.

Before killing anything, potpie checks the process really is a potpie daemon
(process IDs get recycled, so a stale ID could point at something else entirely
— that's why the old code refused in the first place).

### `pot reset` wiped a different pot than everything else showed you

`pot reset` deletes a pot's graph data. It was picking a *different pot* than the
one `pot list` and `potpie status` displayed as current. So the most destructive
command in the tool was aimed at something you were never shown.

The cause turned out to be one bug behind several symptoms. Potpie can pick a pot
in two ways: the pot you last switched to ("active"), or a pot bound to the repo
you're standing in ("repo default"). Commands were working out the right answer,
using it for logging, and then **throwing it away** and letting a second, separate
piece of code work it out again. The two could disagree.

That's also why `potpie status` showed one pot's name next to another pot's
numbers:

| where you ran it | pot it named | claims it reported |
|---|---|---|
| inside the repo | `cli-wide-test` | 0 |
| outside any repo | `cli-wide-test` | 17 |
| `--pot cli-wide-test` | `default` | 19 |

**Now:** the pot is worked out once and used everywhere. `status` tells you which
pot it's talking about, how it picked it, and — if the "active" pot is a different
one — says so outright. `pot list` marks the pot your commands here will actually
use. `pot reset` takes `--pot`, names its target, and warns you if that isn't the
active pot.

---

## Things that were lying to you

**`potpie doctor` described the wrong potpie.** It scanned your `PATH` and
reported on whatever `potpie` it found first — which might not be the one you just
ran. It confidently reported another install's version. With `PATH` trimmed, the
running binary announced it wasn't installed at all. Now it asks itself, and lists
anything else it finds separately.

**`doctor` reported bash's version as your Python version.** uv's launcher script
starts with `#!/bin/sh`, so doctor ran `/bin/sh --version`, got "GNU bash, version
3.2.57", grabbed the first number it saw and printed `python=3.2.57`. It now checks
the thing is actually Python before asking.

**`potpie --version` never printed potpie's version.** It printed the engine's
version and your Python. If you installed 2.0.1 you couldn't find that out from
`--version`.

**`--verbose` promised tracebacks and never printed one.** Now it does. (The
normal message stays vague on purpose — error text can contain your file paths.)

**Diagnostics told you to run `make`.** No Makefile ships in the package, so that
advice only works if you happen to have the source checked out.

---

## Things you couldn't do

**A third of `record`'s types couldn't be written.** `potpie record` accepts 15
types, but 5 of them need an extra field and there was no flag to supply one. Its
own help text gave three examples — two of which were impossible to use:

```
$ potpie record --type decision --summary "..."
error: decision: 'rationale' must be a non-empty string      # no way to pass it
```

Added `--detail key=value`, so all 15 work. Which types need what is now published
by the engine and checked by a test, so the help and the validator can't drift
apart.

**Copy-pasting from `graph catalog` always failed.** The catalog lists views as
`debugging.prior_occurrences`, but every way of passing that back was rejected —
you had to know to split it into `--subgraph debugging --view prior_occurrences`.
Now the name it prints is the name it takes.

**A view was reachable but undiscoverable.** Asking for it said it "requires one
of service, anchor_entity_key" — neither of which is a flag. The answer was
`--scope service:<name>`, findable only by guessing. Errors now name the flag you
type: `--by`, `--scope service:<value>`, and so on.

**Search results printed as empty bullets.** Some results carry their text under a
different key than the printer knew about, so they rendered as `• [resources]` and
nothing else. Four of ten visible results were blank lines while the JSON was full.
Fixed, along with rows that showed up as `[?]`.

**The same fact appeared four times in a row.** Duplicates now collapse — but only
when they're genuinely the same record. Two different records that happen to be
worded identically both survive, and the second is labelled so you can see they're
distinct.

---

## Things that were quietly wrong

**Registering the same repo twice stored two different identities.** `source add
repo .` uses your git remote as the repo's identity, falling back to a filesystem
path if it can't get one. It treated *any* git failure as "no remote" — so a slow
or failing git call silently produced a different identity for the same repo.
Now a real failure says so and stops; only a genuine "this repo has no remote"
falls back to the path.

Re-registering also silently ignored a new `--name` and reported a location that
wasn't what got stored. It now says it reused the existing registration.

**Running under an isolated home still wrote to your real `~/.claude`.**
`CONTEXT_ENGINE_HOME` moves potpie's own state but never moved agent skills, so a
"safe" test run rewrote the tester's actual skill set. `POTPIE_HARNESS_HOME` now
does that, and installs print where the files went.

**`skills status` reported a clean bill of health while ignoring skills it can't
service.** Three `potpie-resource-*` skill folders were installed and being loaded
by the agent; this build doesn't ship them and can't update them, and status said
nothing. It now lists them.

**The backend you configured at setup was ignored.** `potpie setup` records your
choice in `config.json`, but starting the daemon never read it and always used the
default. If you'd configured something else, your graph looked empty.

---

## Three findings that don't apply here

Three items describe features that exist on an unmerged branch, not in 2.0.1 or on
`main` — the tester's machine had leftovers from it. There's nothing to fix in this
lineage: the `potpie resource` command and the links pointing at it, the managed
(remote) host config, and the `knowledge.document_passages` view.

Where those findings had a half that *does* apply here, that half is fixed above:
orphaned-skill reporting, the ignored backend setting, and the blank-row rendering.

---

## Review round

CodeRabbit raised 11 points; 9 were real and are fixed in the second commit. Most
were edge cases around the fixes above — the notable ones:

- `git remote get-url` exits 128 both for "not a repo" *and* for a broken config
  file. I was treating both as "no remote", which reintroduced the exact silent
  fallback I was fixing. Only a real "no repo here" counts now.
- The forced daemon kill matched its command line by substring, so a process
  called `potpie.daemon_helper` would have qualified. It now matches the exact
  command shapes the daemon is started with, and there are tests for the near
  misses.
- Skill version records were keyed only by agent, so two isolated harness homes
  overwrote each other's. Each home gets its own record now; the normal one keeps
  its current filename so existing installs aren't disturbed.

Two I didn't act on, with reasons: `fix` records were said to need a `fix_steps`
field — they don't, a bare `fix` record validates and writes (verified both in the
validator and end to end). And two tests were said to be calling a method without
parentheses — it's a property, and the tests pass.

---

## How this was checked

- 1477 CLI tests, 1184 engine tests — green. 90 of them are new and cover each
  finding.
- Every fix re-run by hand against a throwaway daemon: the upgrade wedge recovers,
  the three pot displays agree, all 15 record types write, every catalog view name
  pastes back in.
- Two `test_daemon_controller` and two `test_product_process_surfaces` tests fail
  in my local sandbox for reasons unrelated to this branch — my dev environment has
  one shared editable install pointing at a checkout that's currently on someone
  else's branch, so a spawned test daemon picks up mismatched halves. They pass on
  a clean checkout.

### DIFF
--- a/potpie/cli/cli_install_status.py
+++ b/potpie/cli/cli_install_status.py
@@ -16,32 +16,51 @@ CLI_TOOL_NAME = "potpie-context-engine"
 CLI_EXECUTABLE = "potpie"
 _UV_TOOL_NAMES = frozenset({"potpie", "potpie-context-engine", "context-engine"})
 
+# Diagnostics that ship with the wheel must only name commands the wheel
+# provides. `make` targets exist in the source checkout, never in an install,
+# so they are added only when the running CLI is an editable repo install.
 _DIAGNOSTIC_COMMANDS = (
+    "potpie doctor",
     "uv tool list",
     "which -a potpie",
-    'head -n 1 "$(command -v potpie)"',
-    "make cli-status",
-    "make cli-install",
 )
+_EDITABLE_DIAGNOSTIC_COMMANDS = ("make cli-status", "make cli-install")
 
 _LOCAL_REINSTALL_HINT = (
-    "Check with `make cli-status` or `potpie doctor`. "
-    "Repo-local reinstall: `make cli-install` "
+    "Check with `potpie doctor`. Repo-local reinstall: `make cli-install` "
     "(builds UI, stops old daemon) — not raw `uv tool install`."
 )
 _PUBLISHED_HINT = (
-    "Check with `make cli-status` or `potpie doctor`. "
-    "Published-package reinstall: `uv tool install potpie` "
-    "(or `pip install potpie`)."
+    "Check with `potpie doctor`. Published-package reinstall: "
+    "`uv tool install --force potpie` (or `pip install --upgrade potpie`)."
 )
 
 
 def collect_cli_install_status() -> dict[str, Any]:
-    """Return install facts for ``potpie doctor`` and ``make cli-status``."""
+    """Return install facts about **the running** ``potpie``.
+
+    Resolution starts at this process (``sys.argv[0]`` / ``sys.executable``)
+    rather than at ``$PATH``: scanning ``$PATH`` first made ``doctor`` report a
+    *different* install's version, and made the running binary declare itself
+    "NOT on PATH" whenever ``$PATH`` did not happen to contain it.
+    """
     paths_on_path = _potpie_paths_on_path()
-    primary_path = paths_on_path[0] if paths_on_path else None
-    python_interpreter = _python_from_script(primary_path) if primary_path else None
-    python_version = _python_version(python_interpreter)
+    running_path = _running_cli_path()
+    # The running entry point is authoritative; PATH entries are context.
+    primary_path = running_path or (paths_on_path[0] if paths_on_path else None)
+    resolved_primary = _realpath(primary_path)
+    other_paths = [
+        path for path in paths_on_path if _realpath(path) != resolved_primary
+    ]
+    is_running = bool(running_path) and primary_path == running_path
+
+    if is_running:
+        # We are the process being diagnosed: ask ourselves, never a shebang.
+        python_interpreter = sys.executable
+        python_version = ".".join(map(str, sys.version_info[:3]))
+    else:
+        python_interpreter = _python_from_script(primary_path)
+        python_version = _python_version(python_interpreter)
 
     listed_uv = _uv_tool_list_status()
     active_uv = _active_uv_tool_from_executable(primary_path)
@@ -51,7 +70,11 @@ def collect_cli_install_status() -> dict[str, Any]:
         and _is_editable_uv_tool(active_uv["tool_root"], str(active_uv["tool_name"]))
     )
 
-    package_version = _package_version_via_interpreter(python_interpreter)
+    package_version = (
+        _installed_package_version()
+        if is_running
+        else _package_version_via_interpreter(python_interpreter)
+    )
     if package_version is None:
         package_version = _installed_package_version()
 
@@ -67,12 +90,24 @@ def collect_cli_install_status() -> dict[str, Any]:
     elif via_uv_tool:
         hint = _PUBLISHED_HINT
 
+    diagnostics = list(_DIAGNOSTIC_COMMANDS)
+    if editable:
+        diagnostics.extend(_EDITABLE_DIAGNOSTIC_COMMANDS)
+
     return {
         "package_name": CLI_TOOL_NAME,
         "package_version": package_version,
+        "product_version": _product_version(),
         "on_path": bool(paths_on_path),
         "paths": paths_on_path,
         "primary_path": primary_path,
+        "running_path": running_path,
+        # The running CLI is on PATH only if some PATH entry resolves to it.
+        "running_on_path": bool(
+            resolved_primary
+            and any(_realpath(path) == resolved_primary for path in paths_on_path)
+        ),
+        "other_paths": other_paths,
         "python_interpreter": python_interpreter,
         "python_version": python_version,
         "runtime_python": sys.executable,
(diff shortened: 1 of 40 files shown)


nndn:
Thanks — worked through all 11. Nine were real and are fixed in `260abf31`; two I'm skipping with reasons below.

**Fixed**

- **`pots.py:480`** — right, an id and a name can select the same pot. Now compares what both resolve to, so `pot reset <name> --pot <id>` is accepted and only genuinely different pots are rejected.
- **`repo_location.py:97`** — good catch, and it reintroduced the exact bug the change was meant to fix. Confirmed your repro: exit 128 covers both `fatal: not a git repository` and `fatal: bad config line 1 in file .git/config`. Only the first is a definitive "no origin"; the second now propagates as `RepoLocationError` instead of silently registering a path identity. Exit 2 is unchanged.
- **`envelope_builder.py:180`** — correct. With no `subject_key`, `object_key` or `predicate` there is nothing to tell two same-worded records apart, so `candidate_key` is now the discriminator in that case. Records that do carry endpoints keep the previous identity, and the reported case (all keys empty) still merges. Tests for both.
- **`workbench_service.py:321`** — agreed, pointing at the pot that just failed was useless. It now says to re-run with the pot the plan was proposed in, and notes `graph propose --json` reports it as `pot_id`. Test updated to assert the failing pot is *not* recommended.
- **`discovery.py:213` / `:257`** — right, `_require_private_regular_file` raises before either classified handler, so an ownership/mode failure carried no recovery action. Both boundaries now translate to `DISCOVERY_UNREADABLE` / `DISCOVERY_CREDENTIAL_UNAVAILABLE` with the restart guidance.
- **`lifecycle.py:95`** — this one matters most since it gates a `SIGTERM`. Substring matching is now exact argv matching: the executable basename must be `potpie-daemon`/`potpied`, or the token straight after `-m` must be exactly `potpie.daemon`. Parametrised tests cover `potpie.daemon_helper`, `grep -rn potpie.daemon /src`, `vim potpie.daemon.py` and `python -c "import potpie.daemon"` — all rejected.
- **`lifecycle.py:543`** — fixed, though not quite as suggested: it now checks the boolean from `_cleanup_runtime_records()` and raises `_raise_cleanup_ownership_conflict(expected_pid=pid)` on failure. I deliberately did **not** pass `expected_pid` into the cleanup call — that is verified against the discovery record, which in this path is precisely the record we cannot read, so passing it would make the removal a silent no-op. Commented at the call site.
- **`composition.py:107`** — right, a non-object root raises `AttributeError`, which isn't caught by `(OSError, ValueError)`. Now checks for a mapping first and falls back to `falkordb_lite`.
- **`skills/targets.py:177`** — agreed, and this was mine to introduce: making `skills_root` vary by `POTPIE_HARNESS_HOME` while the manifest stayed keyed by agent+scope means two roots sharing one `CONTEXT_ENGINE_HOME` overwrite each other's versions. The manifest filename now includes a digest of the resolved root. The default root keeps its historical name so existing installs aren't orphaned — that felt worth the small asymmetry. Regression test uses two roots against one state home.

**Not changing**

- **`context_records.py:50` — `fix` needs `fix_steps`.** It doesn't. `_string_tuple(..., allow_empty=False)` filters empty strings out of the list; it doesn't require a non-empty tuple. `validate_record_payload(record_type="fix", summary="...", details={})` returns `FixRecord(fix_steps=(), ...)`, and `potpie record --type fix --summary "..."` writes successfully end to end against a live daemon. Adding `fix` to the required table would make a currently-working command fail.
- **`test_release_findings_regressions.py:57,67` — `recoverable_by_replacement` is a method.** It's a `@property` (`discovery.py:53`), so the comparisons are against `bool`, not a bound method. Both tests pass.

Suites after the fixes: 1477 CLI, 1184 engine — green, including the new tests above.

